PR #3, opened 2026-05-15
doc: Update README with GPU variant specification

dagi3d:
Added example for specifying GPU variant in README.


Andyyyy64:
LGTM!! ty!!

PR #8, opened 2026-05-15
docs(readme): use uv instead of pipx

max-wittig:
As this project is already using uv and many people as well we can also recommend to use it

See also: https://docs.astral.sh/uv/concepts/tools/

## Testing

- [x] Tests pass (`pytest`)
- [x] Tested on real hardware (if hardware-related)

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,10 +136,10 @@ print(output["choices"][0]["message"]["content"])
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


Andyyyy64:
thanks, agreed 
uv is the better default here. appreciate the pr

PR #28, opened 2026-05-16
Accept Apple-prefixed Silicon GPU aliases

### LINKED ISSUE #23: Accept Apple Silicon simulator aliases with Apple prefix
### Problem

`--gpu "M3 Max"` works, but `--gpu "Apple M3 Max"` is treated as unknown. Users naturally copy the display name, so the simulator should accept both forms.

### Done when

- `whichllm --gpu "Apple M3 Max"` resolves to the same simulated Apple GPU as `--gpu "M3 Max"`.
- The same behavior works for common Apple Silicon names such as M1/M2/M3/M4 Max/Ultra.
- A regression test covers the prefixed alias path.

akmhatey-ai:
## What

Accept Apple-prefixed Apple Silicon names in GPU simulation, so display-name inputs like `Apple M3 Max` resolve the same way as `M3 Max`.

Fixes #23.

## Why

Users often copy the full Apple display name when using `--gpu`. The simulator already has explicit Apple Silicon handling for `M1`-`M4` chip names, but the optional `Apple` prefix caused the lookup to fall through to the discrete GPU database and fail as an unknown GPU.

## Testing

- [x] Tests pass (`pytest`)
- [x] New tests added
- [x] Tested on real hardware

Validation run:

```text
uv run pytest
uv run --with ruff ruff check .
uv run --with ruff ruff format --check .
git diff --check
gitleaks detect --pipe --redact --verbose --no-color
```

Manual repro check:

```text
uv run python -c "from whichllm.hardware.gpu_simulator import create_synthetic_gpu; print(create_synthetic_gpu('Apple M3 Max'))"
```

The manual check now returns `Apple M3 Max (simulated)` with vendor `apple` and the same bandwidth/default memory values as `M3 Max`.

## Notes

This only changes the simulator alias normalization path. It does not change real hardware detection.

### DIFF
--- a/src/whichllm/hardware/gpu_simulator.py
+++ b/src/whichllm/hardware/gpu_simulator.py
@@ -66,10 +66,14 @@ def _lookup_apple_silicon(
     """Match Apple Silicon chip names. Returns (canonical_name, vendor,
     default_vram_gb, bandwidth_gbps) or None.
 
-    Matches are case-insensitive and accept both "M2 Max" and "m2max" forms.
-    Longest match wins so "M2 Ultra" does not get caught by the "M2" entry.
+    Matches are case-insensitive and accept "M2 Max", "m2max", and
+    display-name forms such as "Apple M2 Max". Longest match wins so
+    "M2 Ultra" does not get caught by the "M2" entry.
     """
     compact = re.sub(r"\s+", "", name).lower()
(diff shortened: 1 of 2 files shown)


Andyyyy64 (approved):
Thanks! This is a clean fix, and the tests pass locally.

PR #37, opened 2026-05-16
fix(hardware): stop reporting phantom AMD GPU on Intel-only systems

justindotdevv:
# Fix: Phantom "AMD Graphics" GPU on Intel-only systems

## Summary

`whichllm hardware` reported a non-existent `AMD Graphics` GPU on systems
that only have an Intel iGPU (no AMD/ATI hardware and no `rocm-smi`
installed).

## Root cause

In `src/whichllm/hardware/amd.py`, the lspci fallback gated AMD lines with
a naive substring check:

```python
if not any(vendor in line_lower for vendor in ("amd", "ati")):
    continue
```

Every line emitted by `lspci -mm` contains the string `Intel Corporation`,
and `"ati"` is a substring of `corpor`**`ati`**`on`. The Intel VGA entry
therefore passed the vendor filter, then `_normalize_lspci_name` fell back
to the literal string `"AMD Graphics"` because no real AMD vendor token was
present.

## Fix

Replace the loose substring check with explicit AMD vendor markers
(`advanced micro devices`, `amd/ati`, `[amd]`, `[ati]`,
`ati technologies`) that only appear in genuine AMD/ATI `lspci` rows.

## Reproduction

Hardware under test: Intel i5-12450H, Intel UHD Graphics (Alder Lake-P GT1),
no discrete GPU, no `rocm-smi`.

### Before the fix (faulty output)

```text
$ uv run whichllm hardware

╭─────────────────────────────── Hardware Info ────────────────────────────────╮
│ GPU 0: AMD Graphics — shared memory — BW: N/A                                │
│ GPU 1: Alder Lake-P GT1 [UHD Graphics] — shared memory — BW: N/A             │
│ CPU: 12th Gen Intel(R) Core(TM) i5-12450H — 8 cores (AVX2)                   │
│ RAM: 15.3 GB                                                                 │
│ Disk free: 177.4 GB                                                          │
│ OS: linux                                                                    │
╰──────────────────────────────────────────────────────────────────────────────╯
```

`GPU 0: AMD Graphics` is a false positive — `lspci` confirms only one
display controller:

```text
$ lspci | grep -iE 'vga|3d|display'
00:02.0 VGA compatible controller: Intel Corporation Alder Lake-P GT1 [UHD Graphics] (rev 0c)
```

### After the fix (correct output)

```text
$ uv run whichllm hardware

╭─────────────────────────────── Hardware Info ────────────────────────────────╮
│ GPU 0: Alder Lake-P GT1 [UHD Graphics] — shared memory — BW: N/A             │
│ CPU: 12th Gen Intel(R) Core(TM) i5-12450H — 8 cores (AVX2)                   │
│ RAM: 15.3 GB                                                                 │
│ Disk free: 177.4 GB                                                          │
│ OS: linux                                                                    │
╰──────────────────────────────────────────────────────────────────────────────╯
```

The phantom AMD entry is gone; only the real Intel iGPU is reported.

## Validation

```text
$ uv run pytest tests/ -x -q
........................................................................ [ 50%]
........................................................................ [100%]
144 passed in 0.66s
```

### DIFF
--- a/src/whichllm/hardware/amd.py
+++ b/src/whichllm/hardware/amd.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+import shlex
 import subprocess
 from pathlib import Path
 
@@ -55,17 +56,6 @@ def _make_gpu(
     )
 
 
-def _normalize_lspci_name(line: str) -> str:
-    parts = [p.strip() for p in line.split('"') if p.strip() and p.strip() != "\t"]
-    for i, part in enumerate(parts):
-        if part.lower() in ("advanced micro devices, inc. [amd/ati]", "amd ati"):
-            if i + 1 < len(parts):
-                return parts[i + 1]
-        if "amd" in part.lower() and "ati" in part.lower() and i + 1 < len(parts):
-            return parts[i + 1]
-    return "AMD Graphics"
-
-
 _AMD_VENDOR_MARKERS = (
     "advanced micro devices",
     "amd/ati",
@@ -75,8 +65,9 @@ _AMD_VENDOR_MARKERS = (
 )
 
 
-def _line_has_amd_vendor(line_lower: str) -> bool:
-    return any(marker in line_lower for marker in _AMD_VENDOR_MARKERS)
+def _vendor_is_amd(vendor: str) -> bool:
+    vendor_lower = vendor.lower()
+    return any(marker in vendor_lower for marker in _AMD_VENDOR_MARKERS)
 
 
 def _detect_from_lspci() -> list[str]:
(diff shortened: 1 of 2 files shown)


Andyyyy64:
Thanks for the fix!!!

Could you add a small regression test for the Intel-only `lspci -mm` case? An Intel VGA row with `"Intel Corporation"` should make `detect_amd_gpus()` return no AMD GPUs.

Also, can we check the actual vendor field from `lspci -mm` instead of searching the whole line? That should avoid similar false positives later.

After that, this looks good to merge!

Andyyyy64 (approved):
LGTM, thanks!

PR #49, opened 2026-05-17
fix: treat Apple Silicon as shared-memory in fit detection

### LINKED ISSUE #44: Treat Apple Silicon as shared-memory hardware in fit detection
### Description

Apple Silicon uses unified memory, but whichllm currently does not mark Apple GPUs as `shared_memory=True`.

This is reproducible on the latest published release (`whichllm 0.5.5`) and on current `main` after #42.

`detect_apple_gpu()` stores total physical memory in `GPUInfo.vram_bytes`, but leaves `shared_memory` at the default `False`. The GPU simulator does the same for Apple aliases such as `M1 Pro`, `M1 Max`, and `M3 Max`.

That makes compatibility checking treat Apple Silicon like a discrete GPU with separate CPU RAM. As a result, it can double-count unified memory as:

- GPU memory from `vram_bytes`
- plus additional CPU RAM for partial offload

The speed estimator already handles Apple as unified memory via `gpu.vendor == "apple"`, so the main issue is fit classification, partial-offload warnings, and JSON/display metadata.

### Verified behavior

On `whichllm 0.5.5`:

```json
{
  "name": "Apple M1 Pro (simulated)",
  "vendor": "apple",
  "vram_bytes": 34359738368,
  "memory_bandwidth_gbps": 200.0,
  "shared_memory": false
}
```

On current `main` after #42, the simulator still reports the same `shared_memory: false` value.

Example with current `main`:

```text
M1 Pro 16GB, Q4_K_M
Qwen/Qwen3-30B-A3B   partial_offload   17.08 GiB
warning: ~6% of layers will be offloaded to CPU RAM
```

That is misleading for Apple Silicon because the model is being treated as if it can spill from separate GPU memory into separate CPU RAM, while the machine actually has one unified memory pool.

Another example:

```text
M1 Pro 32GB, Q4_K_M
Qwen/Qwen3-Next-80B-A3B-Instruct   partial_offload   43.97 GiB
```

This has the same issue: unified memory should not be modeled as discrete VRAM plus an extra offload pool.

### Expected behavior

Apple Silicon should be modeled as shared-memory hardware for compatibility and output metadata.

Suggested direction:

- Set `shared_memory=True` in the real Apple detector.
- Set `shared_memory=True` for Apple Silicon simulator aliases.
- Treat `gpu.vendor == "apple"` as shared-memory in compatibility logic as well, so cached/older objects are handled safely.
- Update tests for Apple hardware detection, simulator output, JSON output, and partial-offload fit behavior.

### Notes

This is separate from MoE speed estimation. Speed estimation already has Apple-specific unified-memory handling; this issue is about memory fit classification and display correctness.

devangpratap:
## Summary

Apple Silicon uses unified memory, but whichllm was not setting `shared_memory=True` on Apple GPUs. This caused `check_compatibility` to add `usable_ram` as a separate offload pool on top of `vram_bytes`, even though both represent the same physical memory. A 40 GB model on a 32 GB Mac would report `partial_offload` with 57.6 GB phantom available memory instead of correctly reporting that it cannot fit.

## Changes

**apple.py**: Set `shared_memory=True` in the real Apple detector.

**gpu_simulator.py**: Set `shared_memory=True` for simulated Apple Silicon chips (`--gpu "M2 Max"` etc).

**compatibility.py**: Add `vendor == "apple"` guard alongside `shared_memory` checks on the offload pool and partial offload warning, so cached or older GPUInfo objects without the flag are still handled correctly.

**tests**: Added 7 new tests covering the double-counting fix, full_gpu fit, legacy vendor guard, and simulator shared_memory assertion. All 185 tests pass (178 existing + 7 new).

## Behavior change

Before: Apple GPU with 32 GB unified memory treated as 32 GB VRAM + 25.6 GB offload pool = 57.6 GB total. Models that don't actually fit were classified as `partial_offload` with a misleading "layers will be offloaded to CPU RAM" warning.

After: Apple GPU with 32 GB unified memory treated as 32 GB total. No separate offload pool. Models either fit (`full_gpu`) or they don't (`cpu_only`). Speed estimation is unchanged since `performance.py` already handled Apple via vendor check.

Closes #44

### DIFF
--- a/src/whichllm/engine/compatibility.py
+++ b/src/whichllm/engine/compatibility.py
@@ -57,7 +57,11 @@ def check_compatibility(
             best_gpu_available = gpu_available
 
     vram_available = total_vram if total_vram > 0 else 0
-    offload_ram_available = 0 if best_gpu and best_gpu.shared_memory else usable_ram
+    offload_ram_available = (
+        0
+        if best_gpu and (best_gpu.shared_memory or best_gpu.vendor == "apple")
+        else usable_ram
+    )
 
     # Check compute capability for NVIDIA
     if best_gpu and best_gpu.vendor == "nvidia" and best_gpu.compute_capability:
@@ -94,7 +98,7 @@ def check_compatibility(
             if vram_required > 0
             else 0
         )
-        if best_gpu and best_gpu.shared_memory:
+        if best_gpu and (best_gpu.shared_memory or best_gpu.vendor == "apple"):
             warnings.append("Will use shared system memory")
         else:
             warnings.append(
--- a/src/whichllm/hardware/apple.py
+++ b/src/whichllm/hardware/apple.py
@@ -60,6 +60,7 @@ def detect_apple_gpu() -> list[GPUInfo]:
                 vendor="apple",
                 vram_bytes=unified_memory,  # unified memory
                 memory_bandwidth_gbps=_lookup_bandwidth(chip_name),
+                shared_memory=True,
             )
         ]
     except (KeyError, IndexError, ValueError) as e:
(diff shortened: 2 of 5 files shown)


Andyyyy64 (approved):
thanks, this looks good.

PR #50, opened 2026-05-18
refactor: fetch benchmark scores concurrently via async sources

akx:
## What

* Convert benchmark fetchers to async
* Run benchmark fetchers in parallel
* Move all fetchers to their respective modules
* Move _current_version out of cli.py into whichllm/utils.py so it can be used for user-agent formation

## Why

Makes the "Fetching benchmarks" stage faster.

## Testing

- [x] Tests pass (`pytest`)
  - Not added, just a refactor 
- [x] Tested on real hardware (if hardware-related)
  - Still works fine on my Mac

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import asyncio
 import sys
-from importlib.metadata import PackageNotFoundError, version
 from typing import Optional
 
 import typer
@@ -12,6 +11,7 @@ from rich.console import Console
 
 from whichllm.hardware.types import HardwareInfo
 from whichllm.models.types import GGUFVariant, ModelInfo
+from whichllm.utils import _current_version
 
 app = typer.Typer(
     name="llm-checker",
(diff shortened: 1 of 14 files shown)


Andyyyy64:
thanks for the thoughtful refactor.
this is a really helpful direction, and I appreciate you improving this path.

One behavior change I noticed tho, liveBench and AA Index now raise on HTTP/JSON failures, so the top-level gather skips those sources instead of using the curated fallback. Was that intentional? The current code uses the fallback for those failures, which keeps rankings stable when the live pages are temporarily unavailable.

akx:
> One behavior change I noticed tho, liveBench and AA Index now raise on HTTP/JSON failures, so the top-level gather skips those sources instead of using the curated fallback. Was that intentional?

No, not totally intentional, to be honest :) I'll reinstate the fallback, but on the caller side, so data provenance is less ambiguous.

akx:
Ah, it actually looks like both aa_index and livebench fetches are currently quietly failing since they've changed Next.js versions :)

That's out of scope for this refactor PR though.

Andyyyy64:
thanks, I re-checked this locally. The fallback behavior is fixed, Approved!!!

PR #54, opened 2026-05-19
fix: respect XDG_CACHE_HOME for cache directory

### LINKED ISSUE #46: whichllm should respect XDG_CACHE_HOME
### Problem

`whichllm` uses `~/.cache` directory on unix-like systems, probably because it's part of [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) but it doesn't respect custom (non-default) `$XDG_CACHE_HOME` setting.

### Proposed Solution

`whichllm` should respect `$XDG_CACHE_HOME` setting and only use `~/.cache` if not set.

### Alternatives Considered

_No response_

devangpratap:
## Summary

Fixes #46.

Both `cache.py` and `benchmark.py` hardcode the cache path as `~/.cache/whichllm`. This ignores `$XDG_CACHE_HOME`, which the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/) defines as the preferred user cache location.

**Changes:**
- Add `_cache_dir()` helper in `utils.py` that reads `$XDG_CACHE_HOME` with `~/.cache` fallback
- Replace the hardcoded path in `cache.py` and `benchmark.py` with the shared helper
- Remove now-unused `from pathlib import Path` in both modules

## Test plan

- Added `tests/test_utils.py` with 3 tests:
  - Default (`XDG_CACHE_HOME` unset) resolves to `~/.cache/whichllm`
  - Custom `XDG_CACHE_HOME` is respected
  - Empty `XDG_CACHE_HOME` falls back to default (per XDG spec, empty = unset)
- All 188 existing tests still pass
- `ruff check` and `ruff format --check` clean


Andyyyy64:
This looks good overall. 
One small XDG compliance nit, relative XDG_CACHE_HOME values should be treated as invalid and ignored, so `_cache_dir()` should probably only use it when `Path(base).is_absolute()`.

devangpratap:
Understood. I've added the `Path(base).is_absolute()` guard so relative values are now ignored per the XDG spec. Also added a test for that case. Should be good to go now.

Andyyyy64:
thanks, this looks good now.

PR #55, opened 2026-05-19
fix: inline Livebench data

akx:
## What

Livebench isn't a next.js site (anymore?), so the fetcher always failed.

Instead of having all end users of whichllm fetch and parse the monthly-updated CSV file, this inlines the data (via a separate auxiliary Python script).

## Why

To fix Livebench fetching (as noted in #50's comment)

## Testing

- [x] Tests pass (`pytest`)


Andyyyy64:
thanks, lgtm

PR #60, opened 2026-05-20
feat: support context length shorthand

### LINKED ISSUE #59: Support context length shorthand
### Problem

### Proposed Solution

### Alternatives Considered

_No response_

puneetdixit200:
## Summary
- add parsing for `--context-length` values like `64k` and `128k`
- apply the parser to ranking, plan, upgrade, and run commands
- document shorthand usage in the CLI reference

Closes #59.

## Verification
- `.venv/bin/python -m pytest tests/test_cli.py -q` -> 32 passed
- `.venv/bin/python -m pytest -q` -> 195 passed

## AI-assisted work
This patch was prepared with AI assistance and manually reviewed. The targeted tests were written first, observed failing before implementation, then the full test suite was run locally.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -17,7 +17,7 @@ Common options:
 | Option | Meaning |
 | --- | --- |
 | `--top`, `-n` | Number of ranked models to show. Default: `10` |
-| `--context-length`, `-c` | Context length used for KV cache estimation. Default: `4096` |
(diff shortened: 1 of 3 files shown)


Andyyyy64:
Thanks for working on this.

I’m going to close this one in favor of #64, which covers the same issue and keeps the parsing logic in a reusable CLI parameter type.

I do appreciate the docs updates here. I’ll make sure the final version keeps the CLI docs covered as well.

Issue #26, opened 2026-05-16
Expose benchmark source and confidence in output


Andyyyy64:
### Problem

The output currently shows a compact benchmark status such as `direct`, `estimated`, `self_reported`, or `none`, but it does not expose the underlying source or confidence. That makes it hard to tell whether a recommendation is backed by a direct benchmark, inherited from a base model, interpolated from a family, or only weak evidence.

### Done when

- JSON output includes benchmark evidence details such as source, confidence, and raw/normalized score where available.
- Table output keeps the compact view but gives users an easy way to see why a score is trusted or not.
- Direct, base-model, variant, line-interpolated, self-reported, and missing evidence are distinguishable.
- Tests cover JSON output for at least direct, estimated, self-reported, and none.

Andyyyy64:
This is covered enough by #62 / v0.5.7 to close.

JSON output now includes `benchmark_source` and `benchmark_confidence`, and `docs/cli.md` documents both fields. The table output still keeps the compact benchmark status.

Raw per-source benchmark scores are not exposed separately yet. If that becomes useful, I would rather track it as a smaller follow-up issue instead of keeping this broader one open.

PR #62, opened 2026-05-20
feat: expose benchmark source and confidence in JSON output

samarthpatel24:
### What
  Wire the already-computed `BenchmarkEvidence.source` and `.confidence` through to `CompatibilityResult` and
   JSON output.

  **Changes:**
  - `engine/types.py` — added `benchmark_source` and `benchmark_confidence` fields to `CompatibilityResult`
  - `engine/ranker.py` — copy `bench_evidence.source` and `.confidence` onto the compat result (2 lines)
  - `output/display.py` — include both fields in `display_json()` output dict (2 lines)
  - `docs/cli.md` — documented the new JSON fields
  - `tests/test_ranker.py` — 4 tests covering direct, estimated, self-reported, and none cases
  - `tests/test_cli.py` — 1 test verifying JSON output contains the new fields

  ### Why
  Refs #26 — the ranker already computes 6 granular benchmark source tiers but only exposed a collapsed
  4-value `benchmark_status`. Downstream tools need the full `source` and `confidence` to make informed
  decisions.

  ### Testing
  - All 196 tests pass (`pytest`)
  - `ruff check` and `ruff format --check` clean
  - 5 new tests cover all benchmark evidence tiers

  ### Notes
  Pure additive change — no existing fields modified. New fields default to `"none"` / `0.0` so
  backwards-compatible.


Andyyyy64:
Thanks, this is a useful improvement.

I do not think this should close #26 yet. The PR exposes `benchmark_source` and `benchmark_confidence` in JSON, but #26 also asks for score visibility and an easier table-side way to understand why the evidence is trusted.

Could you change `Closes #26` to `Refs #26` and add the new JSON fields to `docs/cli.md`? With that scope adjusted, this looks good as an incremental JSON output improvement.

samarthpatel24:
Appreciate the feedback — updated the PR to Refs #26 and added the new fields to docs/cli.md. Happy to tackle the table-side visibility in a follow-up PR.

Andyyyy64:
Sorry for the delay here, work has been busy on my end.

Thanks for updating the scope and docs. This works for me as an incremental JSON-output improvement under `Refs #26`; the table-side explanation can stay as follow-up work.

samarthpatel24:
Thanks for the review , I will look into the follow-up changes once I get time off from work.

PR #63, opened 2026-05-20
fix: update CONTRIBUTING.md dev setup to use uv

devangpratap:
## Summary

- CONTRIBUTING.md told contributors to run `pip install -e ".[dev]"`, but pyproject.toml defines dev dependencies under `[dependency-groups]` (PEP 735), not `[project.optional-dependencies]`. The pip command fails.
- Replaced with `uv sync --dev` and `uv run pytest` to match the README and the actual build system.

## Test plan

- [ ] Verify `uv sync --dev` installs dev deps correctly
- [ ] Verify `uv run pytest` runs the test suite


Andyyyy64 (approved):
Thanks, LGTM!

PR #64, opened 2026-05-21
feat: support context length shorthand (64k, 128k)

### LINKED ISSUE #59: Support context length shorthand

devangpratap:
## What

Add shorthand notation for the `--context-length` / `-c` flag across all
commands (`whichllm`, `plan`, `upgrade`, `run`). Users can now write
`whichllm -c 64k` instead of `whichllm -c 64000`.

Supported formats: plain integers (`4096`), k/K suffix (`64k` = 64000),
m/M suffix (`1m` = 1000000), fractional (`1.5k` = 1500). Case-insensitive.

## Why

Closes #59. Converting context lengths to raw integers is annoying when
the values are commonly expressed as 64k, 128k, etc.

## Testing

- [x] Tests pass (`uv run pytest`, 202/202)
- [x] New tests added (12 tests in `test_utils.py` covering valid inputs,
      edge cases, and rejection of bad values)

## Notes

- `ContextLengthType` is a custom `click.ParamType` in `utils.py` that
  handles parsing before typer sees it
- All 4 `--context-length` options in `cli.py` wired via
  `click_type=CONTEXT_LENGTH`
- Help text updated to show shorthand examples
- Plain integer input is fully backwards compatible


Andyyyy64:
Thanks, this is the implementation I’d like to go with for #59.

Could you make one small adjustment before we merge? `64k` currently parses as `64,000`, but for context windows users may expect `64k` to mean `65,536`. I think we should either switch `k` to 1024-based parsing or explicitly document that this is decimal shorthand.

Could you also carry over the CLI docs coverage from #60? After that this looks good to merge.

devangpratap:
Thanks for the feedback! I've pushed the changes:

1. **Switched `k` to 1024-based parsing** — `64k` now parses as `65536` instead of `64000`. Same for `m` (1024²) and `b` (1024³). This matches how LLM context windows are conventionally expressed.
2. **Added CLI docs coverage from #60** — updated `docs/cli.md` with shorthand mentions in the option descriptions and added shorthand examples for all four commands (main, plan, upgrade, run).

All 202 tests pass, ruff clean.

Andyyyy64:
Sorry for the slow follow-up, work has been busy on my end.

This looks good now. The 1024-based shorthand behavior matches how people usually talk about context windows, and the CLI docs coverage is in place.

PR #66, opened 2026-05-26
fix: VRAM display, context warning, GPU entries, and code quality

KunalScriptz:
## Summary

Six bug fixes and code quality improvements:

- **Fix VRAM `0.0` displaying as `"—"` in upgrade table** — `if vram_gb` treated `0.0` as falsy, hiding VRAM for Apple Silicon and integrated GPUs where VRAM reports as 0
- **Fix backward context-length warning condition** — the `model.context_length >= context_length` guard was inverted; the warning only fired when the model *supported* the requested context. Now warns whenever context exceeds 8192 regardless of model support
- **Add missing GPU entries to `constants.py`** — Added `RTX 5060` and `RTX 5070 Ti` to `NVIDIA_COMPUTE_CAPABILITY` (Blackwell, CC 10.0), `RTX 3050` (224 GB/s), `RX 9070` (560 GB/s), and `RX 9070 XT` (640 GB/s) to `GPU_BANDWIDTH`
- **Fix double computation in `aa_index.py` curated fallback** — `_normalize_aa_index()` was called twice per entry in the dict comprehension (once for filter, once for value)
- **Add missing return type annotation** — `_lookup_dbgpu` now explicitly returns `GPUSpecification | None`
- **Deduplicate `_PLAN_GPUS` in `display.py`** — identical 12-entry list was hardcoded in both `display_plan` and `display_plan_json`; extracted to module-level constant

## Test plan

- [x] All 192 tests pass (`uv run pytest`)
- [x] `whichllm --gpu "RTX 3050"` resolves bandwidth
- [x] `whichllm --gpu "RX 9070 XT"` resolves bandwidth  
- [x] `whichllm upgrade "RTX 4090" "RTX 5090"` shows correct VRAM values

PR #67, opened 2026-05-26
feat: support context length shorthand (64k, 128k, 1m)

### LINKED ISSUE #59: Support context length shorthand

KunalScriptz:
## Summary

Adds shorthand parsing for `--context-length` / `-c` so users can write `64k` instead of `65536`.

Supported formats:
- **Plain integers** — `4096`, `8192`, `131072`
- **K-suffix** — `64k` → 65536, `1.5k` → 1536
- **M-suffix** — `1m` → 1048576, `0.25m` → 262144
- **Case-insensitive** — `64K`, `1M`
- **Decimal values** — `1.5k` → 1536

Rejects empty, negative, zero, and malformed values with clear error messages.

Implementation: custom `click.ParamType` subclass (`ContextLengthType`) wired into all four CLI commands (`main`, `plan`, `upgrade`, `run`).

Closes #59.

## Test plan

- [x] 201 tests pass (9 new tests for parsing)
- [ ] `whichllm --context-length 64k` works
- [ ] `whichllm -c 1m` works
- [ ] `whichllm plan "qwen" -c 128k` works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -3,9 +3,11 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import sys
 from typing import Optional
 
+import click
 import typer
 from rich.console import Console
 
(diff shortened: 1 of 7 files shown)


Andyyyy64:
Thanks for taking this on, and sorry for the slow review.

I’m going to close this in favor of #64. It covers the same context-length shorthand feature, keeps the parser in shared utils, includes CLI docs, and already addresses the 1024-based `64k` behavior.

This PR also includes unrelated changes from #66, so it is stacked in a way that makes the context-length change hard to review on its own.

PR #71, opened 2026-05-26
fix: import Exit from typer instead of click in test_cli

### LINKED ISSUE #70: test_cli: Exit import breaks when typer vendors click internally
## Problem

Two tests in `test_cli.py` fail on CI when pip resolves a typer version that vendors click as `typer._click`:

- `test_validate_evidence_rejects_unknown_mode`
- `test_search_model_not_found`

Both tests import `from click.exceptions import Exit` and use `pytest.raises(Exit)`. The source code raises `typer.Exit(code=1)`. When typer vendors click internally, `typer.Exit` resolves to `typer._click.exceptions.Exit`, which is a different class from `click.exceptions.Exit`. The `pytest.raises` block does not catch it and the test fails.

This surfaces on CI because `test.yml` installs with `pip install -e .` (no lockfile), so typer resolution varies across runs. Local runs with `uv sync --dev` pin typer 0.24.1 which does not vendor click, so the tests pass locally.

## Reproduction

Visible in the CI logs for any PR when pip resolves a typer version that vendors click:

```
E           typer._click.exceptions.Exit
FAILED tests/test_cli.py::test_validate_evidence_rejects_unknown_mode
FAILED tests/test_cli.py::test_search_model_not_found
```

## Fix

Change the import in `tests/test_cli.py` from:

```python
from click.exceptions import Exit
```

to:

```python
from typer import Exit
```

This matches what `cli.py` actually raises and works regardless of whether typer vendors click or uses the system package.

devangpratap:
## What

Changes `from click.exceptions import Exit` to `from typer import Exit` in `tests/test_cli.py`.

## Why

Closes #70

`cli.py` raises `typer.Exit(code=1)`. When typer vendors click internally (as `typer._click`), `typer.Exit` becomes `typer._click.exceptions.Exit`. The test imported `Exit` from `click.exceptions` directly, which is a different class in that scenario, so `pytest.raises(Exit)` did not catch it.

This surfaces on CI because `test.yml` installs via `pip install -e .` without a lockfile, so the resolved typer version varies across runs.

## Testing

- [x] Tests pass (`pytest`)
- [x] `ruff check` and `ruff format` clean

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,7 +1,7 @@
 """Tests for CLI helper logic."""
 
 import pytest
-from click.exceptions import Exit
+from typer import Exit
 
 import whichllm.cli as cli_mod
 from whichllm.cli import (


Andyyyy64 (approved):
Thanks for this. Sorry I let it sit for a bit.

This matches the CI failure I'm seeing with Typer's vendored Click import, and the checks are green. I'm going to merge this first since the same failure is blocking a couple of other PRs too.

PR #73, opened 2026-05-28
fix(ranker): demote models that cannot hold requested context

zazzo9039:
## Problem

`rank_models` ranks a model even when its declared max context (`model.context_length`) is below the requested `-c` / `--context-length`. At runtime that model is truncated or rejected, so the top recommendation can be unusable at the asked-for context.

## Fix

- Add a `context_fits` flag on `CompatibilityResult`, set in `check_compatibility`.
- When max context is known and too small, emit a warning and set `context_fits` to `False`.
- Apply a selection penalty in `_family_selection_key` so a model that cannot hold the requested context is demoted below one that can — without removing it.
- If nothing fits, the best available candidate is still returned, with a warning.

When `model.context_length` is unknown, `context_fits` stays `True` (no penalty).

## No behavior change at default context

At `-c 4096`, virtually all models have `context_length >= 4096`, so `context_fits` is `True` for everyone and ranking is unchanged. The penalty only engages when the requested context exceeds a model’s known declared max. The field defaults to `True`.

## Tests

- fit / no-fit / unknown `context_length`
- demotion ordering when two models have similar quality but only one supports the requested context

### DIFF
--- a/src/whichllm/engine/compatibility.py
+++ b/src/whichllm/engine/compatibility.py
@@ -114,7 +114,16 @@ def check_compatibility(
         warnings.append("Insufficient memory (GPU VRAM + RAM) to run this model")
 
     # Context length warning
-    if (
+    context_fits = not (
+        model.context_length is not None
+        and model.context_length < context_length
+    )
+    if not context_fits:
+        warnings.append(
+            f"Model max context {model.context_length} < requested "
+            f"{context_length}; runtime will truncate or reject"
+        )
+    elif (
         context_length > 8192
         and model.context_length
         and model.context_length >= context_length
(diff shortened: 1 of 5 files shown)


Andyyyy64:
Thanks for the PR, and sorry this took a bit.

I reviewed it by merging the PR head into current main locally. The targeted compatibility/ranker tests pass, and the full test suite passes as well.

The approach makes sense to me: known-too-short context gets a warning and a ranking penalty, while unknown context stays unpenalized.

I am going to update the branch and approve Actions so CI can run against current main. Since this touches ranking, I want CI green plus one final look at the penalty before merging.

Andyyyy64 (approved):
CI is green after the format fix, and the final diff still looks scoped. Merging this now.

PR #94, opened 2026-06-08
Fix GGUF benchmark lookup without base_model

### LINKED ISSUE #89: Match community GGUF repos to official benchmark IDs without base_model
Split out from #83.

The current benchmark lookup can inherit scores from `cardData.base_model`. For example, this works:

```text
unsloth/Qwen3.6-27B-GGUF + base_model=Qwen/Qwen3.6-27B
=> source=base_model, score=83.5
```

But if the community GGUF repo does not expose a usable `base_model`, the lookup falls through:

```text
unsloth/Qwen3.6-27B-GGUF + base_model=None
=> source=none
```

That matches the failure mode reported in #83: community GGUF uploads can miss official benchmark scores even when the repo name clearly points at the official model.

Acceptance criteria:

- Add a benchmark lookup path for community GGUF repos when `base_model` is missing.
- Normalize common cases such as:
  - `unsloth/Qwen3.6-27B-GGUF` -> `Qwen3.6-27B` -> `Qwen/Qwen3.6-27B`
  - underscore-separated names such as `Qwen_Qwen3.6-35B-A3B-GGUF`
- Keep the existing parameter compatibility guard, so tiny draft/MTP heads do not inherit scores from much larger base models by accident.
- Add regression tests for both `base_model` present and `base_model` missing.

Parent: #83

cnYui:
## Summary
- match community GGUF repo names to benchmark IDs with the same model name when `cardData.base_model` is missing
- support underscore repo names such as `Qwen_Qwen3.6-35B-A3B-GGUF`
- keep parameter compatibility checks on inherited variant matches

Fixes #89.

## Tests
- `uv run pytest tests/test_benchmark_lookup.py`

## Notes
- `uv run pytest` was run locally on Windows: 231 passed, 1 existing platform-specific failure in `tests/test_utils.py::test_cache_dir_respects_xdg_cache_home` where `_cache_dir()` ignores `XDG_CACHE_HOME` on Windows.

### DIFF
--- a/src/whichllm/models/benchmark.py
+++ b/src/whichllm/models/benchmark.py
@@ -427,6 +427,47 @@ def _generate_candidates(model_id: str) -> list[str]:
     return candidates
 
 
+def _append_unique(candidates: list[str], candidate: str) -> None:
+    if candidate and candidate not in candidates:
+        candidates.append(candidate)
+
+
+def _strip_repo_suffix(model_id: str) -> str:
+    for suffix in _REPO_SUFFIXES:
+        if model_id.endswith(suffix):
+            return model_id[: -len(suffix)]
+    return model_id
+
+
+def _generate_score_name_candidates(
+    model_id: str, scores: dict[str, float]
+) -> list[str]:
+    """Match community repo names to benchmark IDs with the same model name."""
+    stripped = _strip_repo_suffix(model_id)
+    repo_name = stripped.rsplit("/", 1)[-1]
+    model_names = [repo_name]
+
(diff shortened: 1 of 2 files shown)


Andyyyy64:
Thanks @cnYui, this fixes #89. I ran it locally to check: community repo names now resolve to the official benchmark id at `variant` tier (never `direct`), and the ±2× params guard still throws out the draft/MTP case, so the hole closes without wrong-size or self-reported scores sneaking in. Full test suite passed here too (232). Appreciate the regression tests. Merging now.

PR #96, opened 2026-06-09
fix: restore AA live fetch after RSC migration (#87)

### LINKED ISSUE #87: Restore Artificial Analysis live fetch after page format change
Split out from #83.

`fetch_aa_index_scores()` still expects the Artificial Analysis leaderboard HTML to contain a `__NEXT_DATA__` payload. That no longer seems to be true for the current page.

I confirmed the reported failure locally:

```bash
curl -L -s https://artificialanalysis.ai/leaderboards/models
# response does not contain __NEXT_DATA__
```

Current behavior:

```text
AA Index fetch failed, will use fallback: __NEXT_DATA__ payload not found
```

The curated fallback keeps whichllm usable, so this is not a hard crash. But the live AA refresh path is effectively dead until we switch to a current data source or parser.

Acceptance criteria:

- Replace the `__NEXT_DATA__`-only scrape with a working AA data source, if one is available.
- Keep the curated fallback path for schema drift or network failures.
- Add a regression test so a missing `__NEXT_DATA__` payload does not silently look like a successful live fetch.

Parent: #83

samarthpatel24:
Fixes #87

## Approach

AA's RSC push payloads contain `"models":[...]` arrays that are valid JSON after unescaping. We parse those directly instead of regex-matching individual fields — avoids field-order brittleness and a cross-boundary bug where null `intelligenceIndex` models steal the next model's score.

RSC → `__NEXT_DATA__` fallback → `ExtractionFailed` → curated fallback.

## Changes

- `constants.py` — add `_RSC_PUSH_RE` for RSC push chunks
- `aa_index.py` — add `_extract_rsc_pairs()` using JSON parsing; try RSC first in `fetch_aa_index_scores()`
- `tests/test_aa_index.py` — 6 tests

## Tests

- RSC extraction returns correct (name, score) pairs
- Graceful empty on non-RSC HTML
- Deprecated models included
- Null `intelligenceIndex` doesn't steal next model's score
- `ExtractionFailed` when neither format found
- Curated fallback returns valid 0–100 scores
- Live: 491 entries extracted, 0 false matches against JSON ground truth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/), and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- AA name matching now canonicalizes variant-suffixed display names
+  (`(Reasoning)`, `(Non-reasoning)`, `(high)`) so live fetch maps ~46 models
+  instead of ~8.
+- RSC chunk regex now matches any tag number (`[\d+, ...]`) instead of only
+  tag 1, so model data in other RSC segments is no longer silently dropped.
+
 ## [0.5.8] - 2026-06-05
 
 ### Added
--- a/src/whichllm/models/benchmark_sources/aa_index.py
+++ b/src/whichllm/models/benchmark_sources/aa_index.py
@@ -93,6 +93,19 @@ AA_NAME_TO_HF_IDS: dict[str, list[str]] = {
 # normalized, and 8B-class (Qwen3-8B = 30) → 40 normalized. This keeps a
 # strong 8B model competitive with frozen-OLLB 7B scores while still leaving
 # clear headroom for frontier-tier models.
+_PAREN_RE = re.compile(r"\([^)]*\)")
+
+
+def _canonical_name(name: str) -> str:
+    name = _PAREN_RE.sub("", name)
+    name = name.lower().replace("-", " ").replace("_", " ")
+    return re.sub(r"\s+", " ", name).strip()
+
+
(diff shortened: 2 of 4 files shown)


samarthpatel24:
Adopted name canonicalization idea from #97 .

Changes:

- Name canonicalization — _canonical_name() strips parenthetical suffixes ((Reasoning), (Non-reasoning)) and normalizes case/separators. Live mapping coverage goes from ~8 → ~46 models.
- Broadened RSC regex — Now matches all RSC tag numbers, not just tag 1. Safe for our JSON extraction since matches are triple-gated.
- Skipped fallback merge — benchmark.py already handles fallback on extraction failure; merging curated data inside the fetch would pollute the current/frozen tier separation in the scoring pipeline.
- 5 new tests covering canonicalization, distinct model preservation, pipeline integration, and non-tag-1 extraction.

Andyyyy64:
Thanks for this, and sorry about the timing: #97 showed up for the same issue a few hours after yours, and I could only take one.

I compared both against the same saved copy of the live page. Your extractor and #97's pulled the identical 491 (name, score) pairs. Two completely different methods agreeing exactly gave me real confidence in both, and your null-intelligenceIndex test pins down a failure mode worth pinning.

What decided it was the step after extraction. AA reworked their index, so older models now score far below my May snapshot. Returning only live-mapped scores shrinks the AA tier from 72 to 43 models and sends things like Qwen3-8B from 40 to 1.5. #97 overlays live on top of the curated snapshot, so coverage can only grow, and that conservative behavior is what I need until I retune the normalization for the new scale.

Closing in favor of #97, but this was close, and the cross-validation your approach made possible was genuinely useful. Hope to see you on another issue.

samarthpatel24:
Thanks for the detailed review — learned a lot from digging into both approaches. The normalization drift issue makes total sense as the deciding factor. Happy to contribute on the next one

PR #97, opened 2026-06-09
fix(aa-index): scrape App Router RSC payload instead of __NEXT_DATA__

xack20:
# PR: fix(aa-index): scrape App Router RSC payload instead of `__NEXT_DATA__`

**Base:** `Andyyyy64/whichllm:main` · **Head:** `fix/aa-index-rsc-scraper`

---

## Summary

The Artificial Analysis Intelligence Index source has been silently failing on
every run. `artificialanalysis.ai` migrated its leaderboard to the **Next.js
App Router**, which no longer embeds a `<script id="__NEXT_DATA__">` blob — model
data now streams via `self.__next_f.push([n, "…"])` (RSC) chunks. The scraper's
`_NEXT_DATA_RE` regex never matches, so `fetch_aa_index_scores` raises
`ExtractionFailed("__NEXT_DATA__ payload not found")`, and `build_scores()`
logs:

```
AA Index fetch failed, will use fallback: __NEXT_DATA__ payload not found
```

then falls back to the frozen `AA_INDEX_FALLBACK_2026_05_14` snapshot. The CLI
still prints (the failure is caught), but the "current" AA tier is stale on
every invocation.

## Changes

All in `src/whichllm/models/benchmark_sources/aa_index.py`:

1. **Parse the App Router RSC stream.** `_decode_rsc_blob()` concatenates and
   unescapes the `self.__next_f.push([n, "…"])` chunks; `_extract_aa_pairs_from_html()`
   pulls every `{"name", …, "intelligenceIndex"}` record out with a bounded
   regex (the payload is a flat RSC stream, not one parseable JSON document, so
   the middle of the record regex forbids a second `"name":"` to avoid leaking
   across records).
2. **Canonicalize variant-suffixed names.** AA now labels models like
   `"Qwen3 14B (Reasoning)"`, `"GLM-5 (Non-reasoning)"`, `"gpt-oss-20B (high)"`.
   `_canonical_name()` strips parentheticals and normalizes separators/case so
   they map back onto the existing `AA_NAME_TO_HF_IDS` table. This lifts live
   name→HF coverage from **8 → ~46** models without enlarging the table.
3. **Overlay live over the curated fallback.** A successful live fetch now
   merges on top of `get_aa_curated_fallback()` (live wins where both exist),
   so a fetch can only *add* coverage — it can never shrink the AA tier below
   the snapshot. Previously, replacing the ~72-entry snapshot with ~8 exact
   matches would have *regressed* rankings.
4. **Keep `__NEXT_DATA__` as a secondary fallback** in case the site format
   changes again.

## Verification

Against the live page (2026-06-09): live fetch returns **78 merged scores**
(≥72 fallback baseline), with **12 models refreshed** by live data and 6 new
ones not in the snapshot (GLM-4.7, MiniMax-M2, MiMo-V2-Flash, …). The
`AA Index fetch failed` warning no longer appears.

## Tests

Adds `tests/test_aa_index.py` — fully offline (`httpx.MockTransport`), covering:
- name canonicalization (variant + separator stripping),
- RSC chunk decoding and record extraction (incl. the no-leak boundary),
- canonical-name → HF mapping through `fetch_aa_index_scores`,
- the merge-over-fallback coverage guarantee,
- the `ExtractionFailed` path when no records are found.

```
uv run pytest          # 298 passed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/), and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- Artificial Analysis Intelligence Index is fetched live again. The
+  artificialanalysis.ai leaderboard migrated to the Next.js App Router and no
+  longer ships a `__NEXT_DATA__` blob, so every run logged
+  `AA Index fetch failed ... __NEXT_DATA__ payload not found` and silently used
+  the frozen snapshot. The scraper now parses the App Router RSC stream
+  (`self.__next_f.push(...)`), canonicalizes AA's variant-suffixed display
+  names (`(Reasoning)`, `(high)`, ...) for mapping, and overlays live scores on
+  top of the curated fallback so a successful fetch can only add coverage. The
+  legacy `__NEXT_DATA__` path is kept as a secondary fallback.
+
 ## [0.5.8] - 2026-06-05
 
 ### Added
--- a/src/whichllm/models/benchmark_sources/aa_index.py
+++ b/src/whichllm/models/benchmark_sources/aa_index.py
@@ -203,9 +203,7 @@ def _normalize_aa_index(index: float) -> float:
 # ``{"name": …, …, "intelligenceIndex": …}`` record out with a bounded regex
 # (the payload is a flat RSC stream, not a single parseable JSON document).
 
-_RSC_CHUNK_RE = re.compile(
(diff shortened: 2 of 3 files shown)


Andyyyy64:
Merged, thank you. Since #96 landed on the same issue a few hours apart, I tested both against one saved copy of the live page. Both extractors pulled the identical 491 (name, score) pairs, so what decided it was the merge behavior. AA reworked their index and older models now score far below my May snapshot. Without the overlay the AA tier shrinks from 72 to 43 mapped models; with it the tier grows to 78 and nothing drops out.

Verified locally: 298 tests pass, ruff clean, and a live fetch adds 6 models on top of the snapshot. I'll open a follow-up for retuning the normalization bounds to AA's new scale.

PR #103, opened 2026-06-10
feat: replace fixed 80% RAM reserve with bounded-reserve formula

### LINKED ISSUE #51: Make RAM fit checks distinguish planned capacity from current available memory
## Problem

`whichllm` currently estimates usable system RAM as a fixed fraction of total physical RAM:

```python
usable_ram = int(hardware.ram_bytes * 0.80)
```

`hardware.ram_bytes` is currently total physical memory from `psutil.virtual_memory().total`. That makes the fit check easy to explain, but it mixes two different questions:

1. Can this machine realistically run the model if the user frees normal working memory?
2. Can this machine run the model safely right now with current apps open?

This matters especially for shared-memory / unified-memory systems. For AMD shared-memory APUs such as Strix Halo / Ryzen AI MAX / Radeon 890M-class iGPUs, `_gpu_available_memory()` uses the system memory pool when dedicated VRAM is just a small aperture. In those cases, overestimating usable RAM directly overestimates available GPU memory.

## Why the current heuristic is risky

A fixed `total_ram * 0.80` is not an industry-standard fit policy; it is only a simple heuristic. It can fail in both directions:

- **64 GB RAM with 40 GB already used:** `64 GB * 0.80 = 51.2 GB` can overstate what is safe to use right now.
- **1 TB RAM server:** reserving a fixed 20% throws away ~204 GB, which can be overly conservative.
- **16 GB everyday machine:** reserving only 20% may be too little when browsers, IDEs, and background apps are active.

`psutil.virtual_memory().available` is better than total for current-state estimation, but it also should not be treated as fully consumable by an LLM process. It means memory available without swapping, not memory that should be consumed to zero.

## Proposed model

Add current available RAM to hardware detection without changing the meaning of the existing `ram_bytes` field:

```python
@dataclass
class HardwareInfo:
    ram_bytes: int                  # total RAM, existing behavior
    ram_available_bytes: int | None # current available RAM, new
    ...
```

Keep `detect_ram_bytes()` for backward compatibility and add a richer helper:

```python
def detect_memory_info() -> tuple[int, int]:
    vm = psutil.virtual_memory()
    return vm.total, vm.available
```

Then compute two different RAM capacities:

```text
usable_ram_planned:
  total_ram - bounded_total_reserve
  "Can run if the user frees normal working memory"

usable_ram_now:
  min(total_ram - bounded_total_reserve,
      available_ram - bounded_live_reserve)
  "Can run safely right now"
```

A concrete starting point:

```python
def clamp(x: int, lo: int, hi: int) -> int:
    return max(lo, min(hi, x))


def estimate_usable_ram(
    total: int,
    available: int | None,
    *,
    reserve_ratio: float = 0.15,
    min_reserve: int = 4 * GiB,
    max_reserve: int = 32 * GiB,
    live_reserve_ratio: float = 0.05,
    min_live_reserve: int = 2 * GiB,
    max_live_reserve: int = 16 * GiB,
    ignore_current_usage: bool = False,
) -> int:
    total_reserve = clamp(
        int(total * reserve_ratio),
        min_reserve,
        max_reserve,
    )
    total_cap = max(0, total - total_reserve)

    if available is None or ignore_current_usage:
        return total_cap

    live_reserve = clamp(
        int(total * live_reserve_ratio),
        min_live_reserve,
        max_live_reserve,
    )
    live_cap = max(0, available - live_reserve)

    return min(total_cap, live_cap)
```

The exact defaults can be tuned, but the important design point is bounded reserves plus separate planned/current fit estimates.

## Suggested CLI behavior

Add explicit options so users can choose how conservative the recommendation should be:

```bash
whichllm --ram-reserve 8GB
whichllm --ram-utilization 0.85
whichllm --planning
whichllm --strict-now
```

Suggested semantics:

- `--ram-reserve 8GB`: absolute reserve; highest priority and easiest to understand.
- `--ram-utilization 0.85`: cap based on total RAM; document clearly whether this applies to total RAM only.
- `--planning`: ignore current available RAM and use planned capacity.
- `--strict-now`: treat current available RAM as a hard constraint.

`--ram-reserve` is probably the clearest primary knob because `--ram-utilization` can be ambiguous: users may not know whether it means total RAM or currently available RAM.

## Ranking behavior

Avoid making current available RAM an unconditional hard reject by default. Prefer tracking both fit states and warning/penalizing as appropriate:

```text
fit_planned = full_gpu / partial_offload / cpu_only / cannot_run
fit_now     = full_gpu / partial_offload / cpu_only / cannot_run

rank penalty:
  fit_now == fit_planned:
    no penalty
  fit_planned works but fit_now tight:
    small penalty + warning
  fit_planned works but fit_now cannot_run:
    medium penalty unless --planning
  --strict-now:
    hard filter
```

Possible JSON shape:

```json
{
  "ram_total_gb": 64,
  "ram_available_gb": 21.5,
  "usable_ram_planned_gb": 54.4,
  "usable_ram_now_gb": 17.5,
  "ram_policy": {
    "reserve_ratio": 0.15,
    "min_reserve_gb": 4,
    "max_reserve_gb": 32,
    "uses_current_available_ram": true
  },
  "fit_now": "partial_offload_tight",
  "fit_planned": "partial_offload_ok",
  "warnings": [
    "Current available RAM is low; close other applications for this recommendation."
  ]
}
```

## Acceptance criteria

- [ ] `HardwareInfo` includes total RAM and current available RAM without changing the meaning of existing `ram_bytes`.
- [ ] RAM fit estimation distinguishes planned capacity from current available capacity.
- [ ] System RAM reserves are bounded by ratio + min + max instead of a fixed `total * 0.80` policy.
- [ ] Shared-memory / unified-memory fit checks use the improved RAM policy.
- [ ] Default ranking does not over-hard-reject merely because the user has temporary apps open, but it does warn or penalize when current available RAM is tight.
- [ ] `--planning` and `--strict-now` expose the two intended modes clearly.
- [ ] JSON output exposes enough RAM policy data for debugging and issue reports.
- [ ] Tests cover at least: 64 GB with high current usage, 16 GB everyday machine, 1 TB server, and shared-memory APU fit behavior.

## Notes

The key improvement is not just using `available` RAM. The key improvement is making the output distinguish:

- hardware capability / planning estimate
- current-state / right-now estimate

That should make recommendations more accurate and more explainable, especially on Strix Halo, Ryzen AI MAX, Radeon 890M-class systems, Apple Silicon, and CPU-only fallback cases.

samarthpatel24:
## What

Replace the hardcoded `total_ram * 0.80` heuristic with a bounded-reserve function: `total - clamp(total * 0.15, 4 GiB, 32 GiB)`.

- Add `estimate_usable_ram(total: int) -> int` to `hardware/memory.py` with inline constants
- Update both call sites (compatibility checker + CLI auto min-params)
- Add parameterized unit tests + update two existing shared-memory APU assertions

| Total RAM | Old (20% reserve) | New (bounded) |
|-----------|-------------------|---------------|
| 16 GB     | 12.8 GB           | 12.0 GB       |
| 32 GB     | 25.6 GB           | 27.2 GB       |
| 64 GB     | 51.2 GB           | 54.4 GB       |
| 128 GB    | 102.4 GB          | 108.8 GB      |
| 1 TB      | 819.2 GB          | 992.0 GB      |

## Why

The flat 20% reserve wastes too much on large machines (e.g. 200 GB reserved on a 1 TB box) and doesn't reserve enough on small ones. The bounded formula scales the reserve with machine size while capping it at 32 GiB.

Part 1 of #51

## Testing

- [x] Tests pass (`pytest`) — 321 passed
- [x] New tests added — parameterized tests for 16 GB, 32 GB, 64 GB, 128 GB, 1 TB + boundary checks
- [x] Tested on real hardware — 15.3 GB RAM, RTX 4070 Laptop GPU, ranking works end-to-end

## Notes

Part 2 (available-RAM awareness, `--planning`/`--strict-now` flags, dual fit tracking) will follow in a separate PR that closes #51.

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -127,7 +127,9 @@ def _auto_min_params_for_profile(hardware: HardwareInfo, profile: str) -> float 
         return None
     if not hardware.gpus:
         return 2.0  # CPU-only: tiny is the only practical choice
-    usable_ram = int(hardware.ram_bytes * 0.80)
+    from whichllm.hardware.memory import estimate_usable_ram
+
+    usable_ram = estimate_usable_ram(hardware.ram_bytes)
     best_vram_gb = max(
         (usable_ram if g.shared_memory and g.vram_bytes == 0 else g.vram_bytes)
         for g in hardware.gpus
--- a/src/whichllm/engine/compatibility.py
+++ b/src/whichllm/engine/compatibility.py
@@ -8,6 +8,7 @@ from whichllm.constants import VULKAN_ONLY_GPUS
 from whichllm.engine.quantization import estimate_weight_bytes
 from whichllm.engine.types import CompatibilityResult
 from whichllm.engine.vram import estimate_vram
+from whichllm.hardware.memory import estimate_usable_ram
 from whichllm.hardware.types import GPUInfo, HardwareInfo
 from whichllm.models.types import GGUFVariant, ModelInfo
 
@@ -56,8 +57,7 @@ def check_compatibility(
 
     vram_required = estimate_vram(model, variant, context_length)
 
-    # Reserve 20% of RAM for OS and other processes
-    usable_ram = int(hardware.ram_bytes * 0.80)
+    usable_ram = estimate_usable_ram(hardware.ram_bytes)
 
     # Determine best GPU
     best_gpu: GPUInfo | None = None
--- a/src/whichllm/hardware/memory.py
+++ b/src/whichllm/hardware/memory.py
@@ -21,7 +21,7 @@ def estimate_usable_ram(total: int) -> int:
     _GiB = 1024**3
     reserve = int(total * 0.15)
     reserve = max(4 * _GiB, min(reserve, 32 * _GiB))
-    return total - reserve
+    return max(0, total - reserve)
 
 
 def detect_disk_free_bytes(path: str | None = None) -> int:
--- a/src/whichllm/hardware/memory.py
+++ b/src/whichllm/hardware/memory.py
@@ -13,6 +13,17 @@ def detect_ram_bytes() -> int:
     return psutil.virtual_memory().total
 
 
+def estimate_usable_ram(total: int) -> int:
+    """Estimate RAM available for model loading after OS/background reserve.
+
+    Uses a bounded-reserve formula: total - clamp(total * 0.15, 4 GiB, 32 GiB).
+    """
+    _GiB = 1024**3
+    reserve = int(total * 0.15)
+    reserve = max(4 * _GiB, min(reserve, 32 * _GiB))
+    return total - reserve
+
+
 def detect_disk_free_bytes(path: str | None = None) -> int:
     """Get free disk space in bytes at the given path.
 
--- a/tests/test_compatibility.py
+++ b/tests/test_compatibility.py
@@ -1,6 +1,7 @@
 """Tests for compatibility checking."""
 
 from whichllm.engine.compatibility import check_compatibility
+from whichllm.hardware.memory import estimate_usable_ram
 from whichllm.hardware.types import GPUInfo, HardwareInfo
 from whichllm.models.types import GGUFVariant, ModelInfo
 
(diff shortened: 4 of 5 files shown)


samarthpatel24:
Add `max(0, ...)` guard to match the proposed signature in #51 — prevents negative return if total RAM is somehow below the 4 GiB minimum reserve.

Andyyyy64 (approved):
Reviewed this locally. This is the right first step for #51: replace the fixed 80% RAM rule with a bounded reserve, while leaving current available RAM for the follow-up.

Minor note: `estimate_usable_ram` has the negative-return guard, but the tests do not cover a total RAM value below the 4 GiB minimum reserve. I do not think that needs to block this PR. The guard is simple, and the normal machine sizes are covered.

I am going to merge this, then answer the Part 2 questions on #51 separately.

Issue #25, opened 2026-05-16
Improve estimation for context, KV cache, and backend-specific runtime behavior


Andyyyy64:
### Problem

The current estimator is useful, but still too generic for some real workloads. It treats long context mostly as a linear KV-cache cost and does not model things like sliding-window attention, KV cache quantization, batch parallelism, MTP, or backend-specific behavior such as MLX vs llama.cpp vs Vulkan/ROCm.

### Done when

- The estimator has explicit inputs or model metadata for at least sliding-window attention and KV cache quantization.
- Long-context VRAM estimates change according to model architecture instead of one global formula.
- Backend/runtime differences are represented clearly enough that Apple MLX, llama.cpp, Vulkan, ROCm, and CUDA do not all collapse into one rough vendor multiplier.
- Existing estimates remain conservative when metadata is unknown.
- Tests cover dense attention, sliding-window attention, and at least one backend-specific speed path.

Andyyyy64:
I left this too broad for too long. Sliding-window attention support has shipped, but separate runtime models for MLX, llama.cpp, Vulkan, and ROCm need measurements that we do not have.

I'm closing this as not planned in its current form. This does not mean KV-cache quantization or runtime-specific estimates are implemented. A follow-up should start with a concrete workload and comparable measurements.

PR #124, opened 2026-06-19
feat(vram): model sliding-window attention in KV cache estimation

SuperMarioYL:
## Summary

`estimate_kv_cache()` scales the KV cache linearly with the full requested context for **every** model. That over-counts VRAM for sliding-window-attention (SWA) models, whose local-attention layers only ever cache the last `window` tokens. At long context the inflated KV term is large enough to push `engine/ranker.py` (which since #73 demotes models that cannot hold the requested context) into demoting models that actually fit.

Measured on `v0.5.12`:

| Model | Context | KV (before) | KV (after) |
|-------|--------:|------------:|-----------:|
| Gemma-3-27B (1024 window, 1/6 global) | 128K | 12.68 GB | 2.20 GB |
| Mistral-7B-v0.1 (window ignored by runtimes) | 128K | 3.29 GB | 3.29 GB *(unchanged)* |

This PR addresses the sliding-window-attention item of #25.

## Approach

The estimate is reduced **only** for architectures whose mainline runtimes actually honor interleaved SWA — `llama.cpp`'s ISWA path / MLX — namely **Gemma-2, Gemma-3, gpt-oss, and Cohere2**. Everything else keeps the full-context KV figure, so the change can only ever *lower* an estimate where the saving is real and stays conservative otherwise.

- **`ModelInfo`** gains `sliding_window` and `sliding_window_global_ratio` (fraction of layers using full/global attention).
- **`fetcher`** resolves them from authoritative sources first — HF config `sliding_window` / `sliding_window_pattern`, then GGUF metadata architecture — and only then a narrow, boundary-matched, conflict-guarded model-id fallback for config-less GGUF repos. A merge/finetune whose id merely *contains* `gemma-3` (and names another base like `llama`) is **not** given a window.
- **`estimate_kv_cache`** blends the two layer types into an effective context:
  `global_ratio · ctx + (1 − global_ratio) · min(ctx, window)`.

### Why Mistral is deliberately excluded

Mistral-7B-v0.1's config advertises `sliding_window: 4096`, but mainline `llama.cpp` / MLX do not apply SWA for it (and later Mistral releases set `sliding_window: null`). Honoring the declared window would *under-count* VRAM and recommend a model that won't fit — the one direction #25 asks us to avoid. It therefore keeps the dense estimate.

## Conservatism guarantee

`sliding_window = None` reproduces the previous formula byte-for-byte, and the reduction is monotonic (`effective_ctx ≤ ctx` always). A test pins literal KV byte values so a coefficient/formula drift is caught.

## Tests

`uv run pytest` → **412 passed** (`+16`). New coverage: dense (pinned bytes), pure-SWA plateau, hybrid partial growth, never-exceeds-dense, below-window equality, per-arch config resolution (Gemma-3 / gpt-oss), Mistral-not-honored, `use_sliding_window: False` opt-out, Gemma-1 negative, GGUF-metadata + id-hint resolution, merge/boundary false-positive guards, and cache round-trip. `ruff check` / `ruff format --check` clean.

## Scope

Focused on the SWA half of #25. KV-cache quantization and backend-specific speed paths (the other #25 items) are intentionally left for follow-ups.

### DIFF
--- a/src/whichllm/engine/vram.py
+++ b/src/whichllm/engine/vram.py
@@ -19,12 +19,40 @@ _KV_BYTES_PER_BPARAM_PER_KCTX = 3.5 * 1024 * 1024  # 3.5 MB
 _MOE_ATTENTION_PARAM_MULTIPLIER = 4.0
 
 
+def _effective_kv_context(model: ModelInfo, context_length: int) -> float:
+    """Context length that actually contributes to the KV cache.
+
+    For sliding-window-attention (SWA) models, local-attention layers only keep
+    the last ``sliding_window`` tokens, so their KV footprint plateaus once the
+    request exceeds the window. Hybrid models interleave a fraction of global
+    (full-context) layers; that fraction is ``sliding_window_global_ratio``.
+
+    Effective context blends the two layer types::
+
+        global_ratio * ctx + (1 - global_ratio) * min(ctx, window)
+
+    This is only applied for architectures whose mainline runtimes honor SWA
+    (the fetcher leaves ``sliding_window`` ``None`` otherwise), and it can only
+    ever lower the estimate — so a model that does not advertise an honored
+    window keeps the full-context KV figure and stays conservative.
+    """
+    window = model.sliding_window
+    ratio = model.sliding_window_global_ratio
+    if not window or window <= 0 or ratio is None:
+        return float(context_length)
+    ratio = min(max(ratio, 0.0), 1.0)
+    windowed = min(context_length, window)
+    return ratio * context_length + (1.0 - ratio) * windowed
+
+
 def estimate_kv_cache(model: ModelInfo, context_length: int) -> int:
     """Estimate KV cache size in bytes for a given context length.
 
     Dense models: KV ≈ 3 MB × params_b × ctx_k (FP16 K+V across all layers).
     MoE models: scale from active params × an empirical multiplier because
     attention shares across experts.
+    Sliding-window models cap local-layer KV at the window size (see
+    :func:`_effective_kv_context`).
     """
     if model.is_moe and model.parameter_count_active:
         # Active-params × MoE multiplier gives a reasonable proxy for the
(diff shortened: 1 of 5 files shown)


Andyyyy64 (changes requested):
Thanks for the careful work here. The KV formula, cache roundtrip, and no-SWA conservative path look solid in local testing.

The part I do not want to merge yet is the model ID fallback in `_SWA_ID_HINTS`.

For VRAM fit, a false positive is expensive. If we undercount memory, whichllm can recommend a model that does not actually fit. I would keep SWA detection to explicit metadata only: HF config `model_type` / `architectures`, or GGUF architecture metadata. If those are missing, leave `sliding_window` unset and keep the full-context estimate.

Could you remove the ID-name fallback and the related tests? After that, this should be much easier to merge.

SuperMarioYL:
Thanks for the review! Agreed on the false-positive cost for VRAM fit.

I've removed the model-id name fallback entirely — `_SWA_ID_HINTS`, `_CONFLICTING_ARCH_TOKENS`, and `_swa_key_from_id` are gone, along with their tests. SWA detection now relies only on authoritative metadata: HF config `model_type`/`architectures` and GGUF architecture metadata. When none of those name a known SWA family, `sliding_window` is left unset and the estimate stays at the conservative full-context value.

The KV formula, cache roundtrip, and no-SWA conservative path are unchanged. Pushed as d0a0050.

Andyyyy64:
Thanks, this is the change I wanted. Limiting SWA detection to HF config metadata and GGUF architecture metadata keeps the VRAM estimate conservative when the metadata is missing.

The only blocker now is formatting. CI is failing because `src/whichllm/models/fetcher.py` needs one ruff format change.

    uvx ruff format src/whichllm/models/fetcher.py

After that I can re-check and merge.

SuperMarioYL:
Done — ran `uvx ruff format src/whichllm/models/fetcher.py` and pushed (8464618). `ruff format --check .` and `ruff check .` both pass clean repo-wide now. Thanks again for the review.

Andyyyy64 (approved):
Thanks, this fixes the remaining blocker.

I rechecked it locally against current main. The SWA path now relies only on HF config and GGUF architecture metadata, and the missing-metadata path stays conservative. Looks good to merge once the branch is updated or GitHub merges it cleanly.

PR #126, opened 2026-06-20
fix: filter model search by parameter count instead of substring for size tokens

### LINKED ISSUE #107: Substring model search silently returns the wrong size (ask 7B, get 1.7B)
### Description

When I ask `plan` / `snippet` / `run` for a model by size, I sometimes get a model of a completely different size back.

For example, `whichllm snippet "qwen 7b gguf"` gave me a snippet for `Qwen3-1.7B-GGUF` — a 1.7B model — even though I clearly asked for a 7B. Same thing with `plan "gemma 2b"`, which planned hardware for the 12B `gemma-3-12b-it`.

Digging in, the matcher in `_search_model` (`src/whichllm/cli.py`) checks each query word as a plain substring of the model ID:

```python
matches = [m for m in models if all(t in m.id.lower() for t in terms)]
```

The problem is the size token `"7b"` is a substring of `"1.7b"`, `"27b"`, `"17b"`, etc. So a search for `qwen 7b` also matches `Qwen3-1.7B`, `Qwen3.6-27B`, and so on. The tool then just sorts those matches by download count and picks the top one — so whether you get the right size comes down to which repo happens to be most popular, not what you typed.

It bites three commands since they all go through `_search_model`: `plan`, `snippet`, and `run` (with an explicit model name). The main ranking command is fine because it doesn't take a name query.

**Expected:** a size like `7b` should only match actual ~7B models, not 1.7B or 27B.

### Steps to Reproduce

1. Run `uvx whichllm@latest snippet "qwen 7b gguf"`
   → resolves to `MaziyarPanahi/Qwen3-1.7B-GGUF` (1.7B), not a 7B model.
2. Run `uvx whichllm@latest plan "gemma 2b"`
   → resolves to `google/gemma-3-12b-it` (12B), because `"2b"` is inside `"12b"`.
3. Run `uvx whichllm@latest snippet "qwen 3b gguf"`
   → resolves to `Qwen3-30B-A3B-GGUF` (30B), because `"3b"` is inside `"A3B"`.

Minimal proof of the root cause (no models needed):

```bash
python3 -c "print('7b' in 'qwen3-1.7b', '2b' in 'gemma-3-12b')"
# True True   <- both are spurious matches
```

> Note: exact results track live HuggingFace data, but the substring mismatch (ask small, get large / ask large, get small) is consistently reproducible.

### Hardware Info

```shell
GPU 0: NVIDIA GeForce RTX 4070 Laptop GPU — 8.0 GB (CC 8.9, CUDA 13.3) — BW: 256 GB/s
GPU 1: Raptor Lake-S UHD Graphics — shared memory — BW: N/A
CPU: Intel(R) Core(TM) i7-14650HX — 16 cores (AVX2)
RAM: 15.3 GB
Disk free: 353.6 GB
OS: linux
```

### Python Version

3.12.13

### Operating System

Arch Linux

### whichllm Version

0.5.8

devangpratap:
## Summary

Fixes #107. `_search_model` used plain substring matching (`"7b" in model_id`), so querying `"qwen 7b"` could match `Qwen3-1.7B` or `Qwen3-27B` depending on download count.

Size tokens (e.g. `7b`, `0.5b`, `500m`) are now parsed out of the query and matched against `parameter_count` with a [0.7x, 1.5x] tolerance band, same approach the rest of the engine uses (`_has_compatible_parameter_count`, `_params_compatible`). Non-size terms still use substring matching.

## Changes

- `_parse_size_tokens()` splits query into text terms and an optional numeric size
- `_size_compatible()` checks `parameter_count` falls within tolerance of queried size
- 8 regression tests covering 7b/1.7b, 2b/12b, 3b/30B-A3B, millions, decimals, missing metadata, multiple size tokens

## Test plan

- [x] `uv run pytest` passes (404/404)
- [ ] Manual: `whichllm snippet "qwen 7b gguf"` returns a 7B model, not 1.7B
- [ ] Manual: `whichllm plan "gemma 2b"` returns a 2B model, not 12B

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -861,6 +861,48 @@ def _load_models(refresh: bool, include_vision: bool = True):
         sys.exit(1)
 
 
+_SIZE_TOKEN_RE = re.compile(r"^(\d+(?:\.\d+)?)([bm])$", re.IGNORECASE)
+
+
+def _parse_size_tokens(
+    terms: list[str],
+) -> tuple[list[str], float | None]:
+    """Split query terms into non-size terms and an optional size in billions.
+
+    Returns (remaining_terms, size_b) where size_b is None if no size token
+    was found.  Only the first size token is used; subsequent size tokens are
+    kept as plain text terms.  Handles 'b' (billions) and 'm' (millions).
+    """
+    remaining = []
+    size_b: float | None = None
+    for t in terms:
+        m = _SIZE_TOKEN_RE.match(t)
+        if m and size_b is None:
+            value = float(m.group(1))
+            if value <= 0:
+                remaining.append(t)
+                continue
+            unit = m.group(2).lower()
+            size_b = value if unit == "b" else value / 1000.0
+        else:
+            remaining.append(t)
+    return remaining, size_b
+
+
+def _size_compatible(model: ModelInfo, size_b: float) -> bool:
+    """Check whether a model's parameter count is compatible with a query size.
+
+    Uses a tolerance band of [0.7x, 1.5x] to accommodate rounding differences
+    (e.g. a 7B query matching a model with 7.6B actual parameters) while
+    rejecting adjacent model sizes (e.g. 7B vs 4B or 12B).
+    """
+    if model.parameter_count <= 0:
+        return True
+    actual_b = model.parameter_count / 1e9
+    ratio = actual_b / size_b
+    return 0.7 <= ratio <= 1.5
+
+
 def _search_model(models: list, model_name: str):
     """Search for a model by name/ID. Returns single model or exits."""
     query_lower = model_name.lower()
@@ -870,7 +912,13 @@ def _search_model(models: list, model_name: str):
     if not matches:
         matches = [m for m in models if m.id.lower().endswith("/" + query_lower)]
     if not matches:
-        matches = [m for m in models if all(t in m.id.lower() for t in terms)]
+        text_terms, size_b = _parse_size_tokens(terms)
+        matches = [
+            m
+            for m in models
+            if all(t in m.id.lower() for t in text_terms)
+            and (size_b is None or _size_compatible(m, size_b))
+        ]
 
     if not matches:
         console.print(f"[red]No model found matching '{model_name}'.[/]")
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -907,6 +907,7 @@ def _search_model(models: list, model_name: str):
     """Search for a model by name/ID. Returns single model or exits."""
     query_lower = model_name.lower()
     terms = query_lower.split()
+    size_b = None
 
     matches = [m for m in models if m.id.lower() == query_lower]
     if not matches:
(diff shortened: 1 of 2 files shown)


Andyyyy64 (changes requested):
Thanks for the focused fix. The main bug is much better now: `qwen 7b` no longer falls into `1.7B`, and `qwen 3b` no longer falls into `30B-A3B`.

One thing still needs tightening before merge. After the numeric filter, close sizes are still sorted mostly by downloads, so the current behavior is:

`qwen 7b` -> `Qwen/Qwen3-8B`
`qwen 3b` -> `Qwen/Qwen3-4B`

That is safer than the old substring match, but I think we still need the close-size tie-break from #107. Could you prefer the model whose ID size label is closest to the requested size, then use downloads as the fallback?

This should be a small follow-up, not a rewrite.

devangpratap (commented):
Thanks for the feedback. Pushed a commit that sorts matches by closest parameter count first, downloads as tiebreak when a size token is present. Queries without a size token keep the original downloads-only sort. Added 4 tests for the new sort behavior, all 62 pass, ruff clean.

Andyyyy64 (changes requested):
Thanks for pushing the follow-up. This is closer, but one part still does not match the behavior I asked for.

The new tie-break uses `parameter_count`. In local QA, `whichllm plan "qwen 7b"` now selects `AxionML/Qwen3.5-9B-NVFP4` because its parameter count is around 6.6B, even though the visible size label is 9B.

For this search path, I still want the ID size label to be the first tie-break after the numeric filter, then downloads as the fallback.

Could you adjust the sort so it works like this?

- hard-filter by numeric `parameter_count`
- among compatible matches, prefer the repo size label closest to the requested size when one is present
- fall back to `parameter_count` and downloads when there is no usable size label

Please add a regression for `qwen 7b` not choosing a 9B-labeled repo when an 8B or 7B-labeled repo is available.

devangpratap:
> Thanks for pushing the follow-up. This is closer, but one part still does not match the behavior I asked for.
> 
> The new tie-break uses `parameter_count`. In local QA, `whichllm plan "qwen 7b"` now selects `AxionML/Qwen3.5-9B-NVFP4` because its parameter count is around 6.6B, even though the visible size label is 9B.
> 
> For this search path, I still want the ID size label to be the first tie-break after the numeric filter, then downloads as the fallback.
> 
> Could you adjust the sort so it works like this?
> 
> * hard-filter by numeric `parameter_count`
> * among compatible matches, prefer the repo size label closest to the requested size when one is present
> * fall back to `parameter_count` and downloads when there is no usable size label
> 
> Please add a regression for `qwen 7b` not choosing a 9B-labeled repo when an 8B or 7B-labeled repo is available.

Understood. Working on it. Will commit by EOD EST.

devangpratap:
Pushed the fix — sort now uses ID label distance (via `_extract_id_size_b`) as the primary tie-break instead of `parameter_count`. Added a regression test for `qwen 7b` not picking 9B-labeled over 8B, plus a parametrized unit test for the extractor. 70 tests pass, ruff clean.

Andyyyy64 (approved):
Thanks for the follow-up. This now matches the behavior I was asking for.

The size token is handled as a numeric filter first, and the tie-break now prefers the closest size label in the repo ID before falling back to parameter count and downloads. I also checked the `qwen 7b`, `qwen 3b`, `gemma 2b`, and `500m` cases locally against current `main`.

Looks good to merge.

PR #133, opened 2026-06-24
Igpu hardware overrides

### LINKED ISSUE #132: Allow manual override of RAM bandwidth and VRAM size for iGPU environments
### Problem

When running whichllm, I get results where models that are clearly impossible to run on my system appear near the top of the recommendations.  
Image

### Proposed Solution

It seems that the tool is overestimating the available RAM bandwidth and VRAM capacity on iGPU-based systems.

Would it be possible to add an option to manually override these values (e.g., RAM bandwidth, usable VRAM size) so that the ranking becomes more realistic for iGPU users?

This would greatly help users whose actual memory bandwidth or GPU memory allocation differs from the automatically detected values.

### Alternatives Considered

_No response_

hannibal-lee:
## What

  Add manual hardware override options for detected GPU/iGPU environments.

  - Allow `--vram` to override detected GPU usable VRAM, not only simulated GPUs
  - Add `--bandwidth` / `--ram-bandwidth` to override GPU/RAM bandwidth estimates
  - Add `--gpu-index` for selecting which detected GPU to override on multi-GPU systems
  - Preserve existing `--gpu ... --vram ...` simulation behavior
  - Document the new iGPU/unified-memory override workflow

  ## Why

  Fixes #132.

  Some iGPU and unified-memory systems can report misleading available GPU memory or memory bandwidth. This can cause whichllm to rank models that appear runnable on paper but are unrealistic on the
  actual machine.

  Manual overrides let users provide their real usable VRAM and RAM bandwidth so ranking and speed estimates are more accurate.

  ## Testing

  - [x] Tests pass (`uv run --no-dev --with pytest pytest`)
  - [x] New tests added
  - [ ] Tested on real hardware

  ## Notes

  This branch is based on latest `origin/main`.

  For example:

  ```bash
  whichllm --vram 8 --ram-bandwidth 68
  whichllm --gpu-index 1 --vram 8 --bandwidth 68

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,9 @@ whichllm
 # Pretend you have a specific GPU
 whichllm --gpu "RTX 4090"
 
+# Override detected iGPU/unified-memory limits
+whichllm --vram 8 --ram-bandwidth 68
+
 # Only show models that fit fully in GPU VRAM
 whichllm --gpu-only
 whichllm --fit gpu
@@ -232,6 +235,8 @@ whichllm --gpu "RTX 4090"
 whichllm --gpu "RTX 5090"
 # Specify variant
 whichllm --gpu "RTX 5060 16"
+# Override detected iGPU/unified-memory limits
+whichllm --vram 8 --ram-bandwidth 68
 # Simulate multiple GPUs
 whichllm --gpu "2x RTX 4090"
 whichllm --gpu "RTX 4090" --gpu "RTX 3090"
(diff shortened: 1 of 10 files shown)


Andyyyy64 (approved):
Thanks for the PR. I checked this against current `main` after #126, and the shape looks good to me.

The detected-GPU override path stays separate from simulation, `--gpu-index` handles multi-GPU detected hardware, and the shared-memory override prevents tiny iGPU placeholder VRAM from being expanded back to system RAM. I also checked the CLI help, hardware output, ranking JSON, `--gpu-only`, and reject cases like `--cpu-only --vram` and `--gpu ... --gpu-index`.

Looks good to merge.

Issue #101, opened 2026-06-10
Retune AA index normalization and refresh snapshot for AA's reworked index


Andyyyy64:
AA reworked their Intelligence Index, and older models now score far lower than the 2026-05-14 snapshot assumed. Live raw for Qwen3-8B maps to 1.5 normalized vs 40.0 in the snapshot. #97 merged with a conservative max-merge overlay, so live downgrades never propagate. To make live data authoritative again:

- Re-derive `_AA_INDEX_MIN` / `_AA_INDEX_MAX` from the current score distribution
- Regenerate the AA fallback snapshot from a fresh scrape on the new scale
- Switch the overlay from max-merge to live-wins

This sits at the core of ranking accuracy, so I'll take it myself.

devangpratap:
@Andyyyy64 if you're not working on this, I'd like to do it.

Andyyyy64:
Thanks. You can take a small first slice if you are still interested.

Please do not change the ranking merge policy in the same PR. The safest first PR would be:

```text
1. refresh the AA fallback snapshot
2. update the normalization bounds from the current AA scale
3. add tests showing the new normalized values for a few known models
```

I want to review the live-wins change separately because that can move rankings more broadly.

Andyyyy64:
#139 is merged and released in v0.5.15.

That covers the first slice here. It retunes the AA bounds, refreshes the fallback snapshot on the reworked AA scale, and adds tests for the new normalized values.

I am keeping this issue open for the remaining live-wins merge policy change. That part can move rankings more broadly, so I still want it in a separate PR.

Andyyyy64:
The live-score update is ready in #173, and all CI checks pass. It now accepts decreases and zero values instead of retaining a higher stored AA score. The PR is awaiting review; this has not shipped yet.

PR #138, opened 2026-06-29
Resolve GGUF artifact links

hannibal-lee:
## What

  - Resolve ranked GGUF recommendations to their actual downloadable artifact repo/file.
  - Add artifact metadata to ranking results and JSON output.
  - Update Rich/Markdown output so model links can point to the real GGUF repo while preserving the recommended base model ID.

  ## Why

  Fixes #137.

  Some recommendations are scored as an official base model plus a GGUF quantization, but the actual GGUF files may live in a third-party repo. The output previously linked only to the base model repo, so
  users could click a `Q3_K_M` recommendation and land on a repo that does not contain that GGUF artifact.

  ## Testing

  - [ ] Tests pass (`pytest`)
  - [x] New tests added (if applicable)
  - [ ] Tested on real hardware (if hardware-related)

  Manual verification:
  - Ran `git diff --check`.
  - Ran `python -m compileall src/whichllm tests/test_artifacts.py tests/test_markdown_output.py tests/test_cli.py`.
  - Could not run full `pytest` locally because `uv run pytest` failed while building `pyarrow==23.0.1`, and the local Python environments do not have `pytest` installed.

  ## Notes

  This keeps ranking based on the recommended/base model, but exposes the resolved GGUF artifact repo and filename for output/download surfaces.

PR #139, opened 2026-06-29
fix: retune AA index bounds and refresh fallback snapshot for the reworked AA scale

devangpratap:
Part of #101 (partial - this is the first slice you scoped, not the full issue, so please keep the issue open).

## What this does

AA reworked their Intelligence Index and the open-weights raw scores compressed a lot, so the old 12.5/56.2 bounds and the 2026-05-14 snapshot no longer line up with the live data. On the old bounds, live raw for Qwen3-8B (around 7.4 now) normalizes to roughly 0, while the curated snapshot still reported a raw value that normalized to 40 for the same model.

1. Re-derived the normalization bounds from the current distribution: `_AA_INDEX_MIN = -19.4`, `_AA_INDEX_MAX = 47.6`. I kept the same two-point calibration the code already used: the top open model (DeepSeek-V4-Pro, raw 44.3) maps to 95, and the 8B class (Qwen3-8B, raw 7.4) maps to 40. On the compressed scale that fit puts the floor below zero. Live AA raw values are always positive and the output is clamped to 0 to 100, so the negative floor only sets where the curve sits.

2. Refreshed the curated fallback from a fresh 2026-06-29 scrape. Models AA currently tracks carry their real new-scale raw values. The peer entries that AA does not track keep their previous normalized score: their raw value is set so it reproduces that score under the new bounds, so the bounds change does not move them.

3. Added tests asserting the new normalized values for known models (top open model around 95, Qwen3-8B at 40, clamp behavior, and a few fallback values).

## What this deliberately does not do

I did not touch the overlay merge policy in `fetch_aa_index_scores`. It is still max-merge: a live score only replaces a snapshot score when the live value is higher. The switch to live-wins is left for a separate PR, as you asked. The snapshot is still stored as raw AA values normalized on read, same as before.

## One thing worth a look

On the reworked scale, several peer entries (models AA does not track) fall below the index floor and now read as negative raw values. They normalize correctly because the output is clamped, but storing negative numbers in a raw-index table is a bit odd. If you would rather, the snapshot could instead store already-normalized 0 to 100 values, which drops the negatives and means a future bounds retune cannot silently shift the fallback. I left that out of this PR since it is more than the issue asks for, but happy to do it in a follow-up if you want it.

## Note on the README

The "What can I run?" example table is marked as a 2026-05 snapshot of illustrative scores. This retune shifts those numbers slightly, but I left the README untouched since refreshing it is outside this issue and needs a live run. Flagging it so you can update it whenever you next regenerate that example.

## Testing

- `uv run pytest`: 456 passed
- `ruff check .` and `ruff format --check .`: clean

### DIFF
--- a/src/whichllm/models/benchmark_sources/aa_index.py
+++ b/src/whichllm/models/benchmark_sources/aa_index.py
@@ -87,106 +87,121 @@ AA_NAME_TO_HF_IDS: dict[str, list[str]] = {
     ],
 }
 
-# AA Intelligence Index typically ranges 12..56 for open-weights models
-# (2026-Q2 frontier tops out at 54; 8B-class lands ~30). The 12..56 window
-# is anchored by a two-point fit: top open frontier (Kimi K2.6 = 54) → 95
-# normalized, and 8B-class (Qwen3-8B = 30) → 40 normalized. This keeps a
-# strong 8B model competitive with frozen-OLLB 7B scores while still leaving
-# clear headroom for frontier-tier models.
-_AA_INDEX_MIN = 12.5
-_AA_INDEX_MAX = 56.2
+# Bounds used to normalize raw AA index values onto the 0-100 scale the rest
+# of the ranking system uses. AA reworked their Intelligence Index in
+# 2026-Q2 and the open-weights distribution compressed sharply (top open model
+# ~44, 8B-class ~7, weakest mapped repos ~3). The window is anchored by the
+# same two-point fit as before: top open frontier (DeepSeek-V4-Pro = 44.3) → 95
+# normalized, and 8B-class (Qwen3-8B = 7.4) → 40 normalized. This keeps a strong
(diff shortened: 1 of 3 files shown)


Andyyyy64 (approved):
Thanks for splitting this out from #101. I checked the retuned bounds, the refreshed fallback, and the live AA fetch path locally.

The calibration does what the issue asked for. DeepSeek V4 Pro lands around 95, Qwen3 8B lands around 40, and the negative raw peer entries still normalize to positive scores instead of dropping out of the fallback. The live fetch is also still parsing current AA data and merges over the fallback correctly.

One small note for a follow-up. The comment around the negative peer raw values says they are kept correct by the clamp. In practice they are correct because the retuned floor is below those values, so they normalize above zero. I do not think that needs to block this PR.

I also tested this merged with current `main` after #138 and #142. This is good to merge.

Issue #41, opened 2026-05-16
Unslop the codebase: pay down ai slop in oversized core modules


Andyyyy64:
## Problem

The codebase works, but several core modules have accumulated too much responsibility in single files. This is intentionally blunt: parts of the implementation now read like ai slop, where CLI orchestration, external I/O, domain rules, scoring heuristics, rendering, and curated data all sit together.

That makes future ranking changes riskier than they need to be. Small edits to benchmark logic, model fetching, output formatting, or CLI behavior can have a wide blast radius because the boundaries are not clean.

## Current hotspots

Measured with `wc -l`:

```text
1087  src/whichllm/cli.py
 880  src/whichllm/models/fetcher.py
 835  src/whichllm/output/display.py
 803  src/whichllm/engine/ranker.py
 751  src/whichllm/models/benchmark.py
 404  src/whichllm/constants.py
```

Tests with similar size pressure:

```text
574  tests/test_cli.py
553  tests/test_ranker.py
500  tests/test_p1_p3_regressions.py
459  tests/test_r3_regressions.py
```

## Suggested direction

Refactor toward clearer responsibility boundaries:

- Keep `cli.py` as a thin Typer layer: option parsing, validation, and command dispatch only.
- Add application/use-case modules for command flows, e.g. `recommend`, `plan`, `upgrade`, `run`, `snippet`, and `hardware`.
- Split `ranker.py` into candidate generation, filtering, scoring, family deduplication, and ranking orchestration.
- Split `benchmark.py` into benchmark fetching, source normalization, score lookup, and evidence resolution.
- Split `fetcher.py` into HuggingFace client calls, model parsing, GGUF extraction, MoE/param overrides, and serialization.
- Split `display.py` by output surface: ranking, plan, upgrade, JSON, and shared formatting helpers.
- Move large curated registries out of `constants.py` where practical, or at least separate GPU, quantization, lineage, and model override data.

## Acceptance criteria

- Core runtime modules are below 400 lines where practical, excluding intentionally data-only files.
- CLI command functions do not contain model loading, benchmark fetching, ranking, and rendering logic inline.
- Ranking behavior is preserved by existing tests.
- The current regression suite still passes.
- New module boundaries make it possible to modify benchmark evidence, scoring, or Rich output without touching unrelated command code.

## Non-goals

- Do not change ranking behavior as part of the first cleanup unless required by extraction.
- Do not reformat the entire codebase mechanically.
[...]

PR #140, opened 2026-06-29
refactor(fetcher): split HuggingFace model fetching modules (#41)

RaghavRD:
Part of #41.

## What

Splits the oversized `src/whichllm/models/fetcher.py` into focused modules by responsibility:

| Module | Purpose | Lines |
|---|---|---:|
| `models/fetcher.py` | compatibility shim / legacy import surface | 110 |
| `models/hf.py` | HuggingFace API orchestration and endpoint handling | 293 |
| `models/parser.py` | HF payload -> `ModelInfo` parsing | 295 |
| `models/parameters.py` | parameter-count, MoE active-param, curated override normalization | 218 |
| `models/gguf.py` | GGUF quant extraction, size estimation, split-file aggregation | 84 |
| `models/sliding_window.py` | sliding-window-attention metadata resolution | 95 |
| `models/serialization.py` | cache serialization/deserialization helpers | 94 |

`fetcher.py` remains as a re-export shim so existing imports like:

```python
from whichllm.models.fetcher import fetch_models, _parse_model, _extract_quant_type


Andyyyy64 (approved):
Thanks for taking this on. I reviewed the split against the old `fetcher.py` surface and checked the current import paths that tests and downstream users are likely to rely on.

The shape looks right to me. `fetcher.py` stays as the import surface, while HF orchestration, parsing, parameter handling, GGUF parsing, SWA metadata, and cache serialization now live in separate modules.

I also tested this merged with current `main` after #138, #139, and #142. The artifact resolution path still works, the fetcher tests still pass, and the full suite passes on the combined tree.

Good to merge.

PR #142, opened 2026-06-30
fix(cli): validate ranking flags to avoid silently wrong output

SuperMarioYL:
## What

The ranking/filter flags `--top`, `--min-speed`, and `--min-params` were never validated, so invalid values silently produced misleading output instead of a clear error — unlike `--vram`, `--bandwidth`, and `--gpu-index`, which already fail fast.

The motivating case is `--top`, which flows straight into `results[:top_n]` in `rank_models`:

- `--top 0` → `results[:0]` → **no recommendations are shown at all**, with no explanation.
- `--top -5` → `results[:-5]` → slices from the end, **silently returning a truncated subset** (the lowest-ranked entries dropped) instead of the count the user asked for.

Negative `--min-speed` / `--min-params` thresholds are meaningless (they turn the filter into a no-op), so they are rejected too.

## Changes

- Add `_validate_ranking_flags()` in `cli.py` (`--top` ≥ 1, `--min-speed` ≥ 0, `--min-params` ≥ 0), following the existing `_validate_gpu_flags` style, and wire it into both the default `recommend` flow and the `upgrade` command (which also exposes `--top`).
- Defense in depth: clamp `top_n` in `rank_models` to `results[: max(top_n, 0)]` so direct callers of this public helper never get a truncated ranking from a stray negative count.
- Tests for each rejected flag and the clamp behavior.

## Notes

- `--context-length` already raises on non-positive values via `parse_context_length`, so it is intentionally left unchanged.
- `ruff check` and `ruff format --check` are clean; the full test suite passes locally (the only failure is the pre-existing `test_run_exits_gracefully` smoke test, which needs a working `uv`/network and fails identically on an unmodified checkout — unrelated to this change).

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -91,6 +91,32 @@ def _validate_output_flags(json_output: bool, markdown_output: bool) -> None:
         raise typer.Exit(code=1)
 
 
+def _validate_ranking_flags(
+    top: int,
+    min_speed: float | None,
+    min_params: float | None,
+) -> None:
+    """Validate ranking/filter flags that otherwise silently distort output.
+
+    Without these guards a non-positive ``--top`` reaches ``results[:top_n]`` in
+    :func:`whichllm.engine.ranker.rank_models`: ``--top 0`` returns no
+    recommendations at all, and a negative value slices from the end
+    (``results[:-5]``), silently returning a truncated, unrequested subset
+    instead of the count the user asked for. Negative ``--min-speed`` /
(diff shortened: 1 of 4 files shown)


Andyyyy64 (approved):
Thanks for the focused fix. I checked the CLI path and the direct `rank_models` guard locally.

`--top 0`, `--top=-5`, negative `--min-speed`, and negative `--min-params` now fail with a clear error instead of producing empty or sliced rankings. I also checked `upgrade --top 0`.

I merged this branch with current `main` locally after #138 and the combined tree still passes the full suite. Looks good to merge.

PR #146, opened 2026-07-16
feat: add Jetson Xavier unified memory detection

hernandez42:
## Summary

Add Xavier family markers to `_NVIDIA_UNIFIED_MEMORY_MARKERS` so `whichllm` correctly detects NVIDIA GPUs on Jetson Xavier platforms that use unified memory.

## Changes

- **File:** `src/whichllm/hardware/nvidia.py`
- Added `"JETSON AGX XAVIER"`, `"JETSON XAVIER NX"`, and `"XAVIER"` to the unified memory marker tuple

## Problem

Jetson AGX Xavier and Xavier NX devices were not detected as having a GPU because their hardware names weren't in the unified memory detection list. `whichllm` would report no GPU found even though the device has a Maxwell GPU with unified memory.

## Solution

Extended the marker list to cover all Xavier variants, matching the existing pattern used for Orin (GB10, DGX SPARK).

## Testing

On a Jetson AGX Xavier or Xavier NX device:
```bash
pip install whichllm
whichllm --detect
```

Should now show the GPU model and unified memory size instead of 'No GPU detected'.


Andyyyy64 (changes requested):
really sorry this took me so long to get to, and thanks for the contribution.

Xavier does use unified system memory, so the direction makes sense. I found a few things we need to clear up before merging:

- `ruff format --check .` currently fails on `src/whichllm/hardware/nvidia.py`.
- the marker is only consulted after NVML or `nvidia-smi` returns a non-empty GPU name. so if the real failure is genuinely “No GPU detected” because neither path enumerates a GPU row, adding `XAVIER` here won’t fix that case. could you include the actual Xavier output or error from NVML / `nvidia-smi`, and add a regression test that reproduces that exact path?
- the architecture comments are mixed up: Jetson AGX Xavier uses a Volta GPU, not Maxwell, while GB10 / DGX Spark are Grace Blackwell systems rather than Jetson Orin. NVIDIA’s specs confirm both: [Jetson AGX Xavier](https://developer.nvidia.com/sites/default/files/akamai/Jetson_AGX_Orin_Developer_Kit_RG_0.pdf) and [DGX Spark](https://www.nvidia.com/en-sg/products/workstations/dgx-spark/).

once we have the real device name/output captured in a test and the comments/formatting corrected, this should be straightforward to re-review.

PR #147, opened 2026-07-20
fix(cli): escape metadata in generated scripts

hannibal-lee:
## Summary

  - encode Hugging Face model IDs, GGUF filenames, and quantization types as
    Python literals before embedding them in generated scripts
  - apply the same escaping to both `whichllm run` and `whichllm snippet`
  - add regression tests for metadata containing quotes and Python-like text

  ## Why

  `whichllm run` generates a temporary Python script from Hugging Face metadata
  and executes it. Direct string interpolation could produce invalid Python or
  allow specially crafted metadata to alter the generated code structure.

  Using `repr()` through the `!r` conversion preserves the original value while
  emitting a correctly escaped Python string literal. Normal model IDs and
  filenames keep the same runtime behavior.

  ## Testing

  - `uv run --no-dev --with pytest pytest -q` (`468 passed`)
  - `uvx ruff check .`
  - `uvx ruff format --check .`

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -1133,12 +1133,15 @@ def _generate_chat_script(model, variant, context_length: int, cpu_only: bool) -
     """Generate a self-contained Python chat script for any model type."""
     if variant:
         n_gpu = 0 if cpu_only else -1
-        return f'''\
+        return f"""\
 from huggingface_hub import hf_hub_download
 from llama_cpp import Llama
 
-print("Downloading {model.id} ({variant.quant_type})...")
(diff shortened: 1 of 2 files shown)


Andyyyy64 (approved):
Hey, really sorry this took me so long to get to. Thanks for waiting!

I finally went through this locally. With a safe AST-only payload, main produced an extra top-level assignment from the model ID, while this branch kept the same payload as the `model_id` string literal.

I also checked both the GGUF and Transformers paths with quotes and newlines. all 468 tests pass, along with Ruff 0.15.22 check and format, and `compileall`.

nothing blocking on my side. Thanks again for fixing this!

PR #151, opened 2026-08-05
ci: use locked development tools in required checks

### LINKED ISSUE #150: ci: use locked development tools in required checks
The CI setup has been drifting for a while, and I should've caught this sooner. Sorry it took me so long to clean it up.

## Problem

The lint and test workflows install development tools directly from the package index instead of using the committed `uv.lock`.

The lint run for PR #147 installed Ruff 0.15.22 and passed. On 2026-07-29, Ruff 0.16.0 reports 100 findings against the same main branch. This means an unchanged commit can get a different required-check result depending on when CI runs.

The test workflow has the same problem with unbounded pytest and pyarrow installations.

## Proposed change

- Declare Ruff in the development dependency group and record it in `uv.lock`.
- Make the lint and test workflows install development dependencies from the lockfile.
- Run Ruff and pytest through `uv`.
- Keep the contributor commands consistent with CI.

## Acceptance criteria

- A fresh locked development sync succeeds on Python 3.11, 3.12, and 3.13.
- Locked Ruff check and format commands pass.
- The full test suite passes with locked dependencies.
- CI no longer installs unbounded Ruff, pytest, or pyarrow packages directly.
- `uv lock --check` succeeds after validation.

## Out of scope

- Fixing or adopting the new Ruff 0.16 findings.
- Changing runtime dependency ranges.
- Changing the PyPI publishing workflow.
- Reformatting or refactoring application code.

Andyyyy64:
Closes #150

really sorry this took me so long to get over the line. The fix was already sitting on a branch, but I left it without a PR for a week.

## What changed

- declare uv 0.11.33 as the repository-required version
- add Ruff 0.15.22 to the development dependencies and `uv.lock`
- make lint and test workflows install from the committed lockfile
- run Ruff and pytest through `uv`
- keep the contributor commands in sync with CI

## Why

The workflows installed Ruff, pytest, and pyarrow directly from the package index without fixed versions. That meant an unchanged commit could get different required-check results depending on when CI ran.

With this change, local development and GitHub Actions resolve the same tools from the repository configuration and lockfile. Tool upgrades now need an explicit code change.

## Validation

- fresh locked sync on Python 3.11, 3.12, and 3.13
- 467 tests passed on each Python version
- locked Ruff check passed
- Ruff format check passed with 98 files unchanged
- `uv lock --check` passed
- both workflow files parsed as YAML
- setup-uv SHA matches the v9.0.0 tag

The first remote run exposed an existing network-dependent smoke test. I split it into deterministic uv-missing and model-missing cases, then reran the full matrix. The follow-up run passed all four required checks.


Andyyyy64:
hey @hannibal-lee, sorry to pull you back into this after I've been so slow on the recent PR queue. This locks the CI tools used by the test and lint checks, and the current run is green.

Would you mind giving it a quick look when you have a chance? no rush, and thanks.

Issue #36, opened 2026-05-16
Improve Strix Halo / Ryzen AI MAX modeling beyond v0.5.4 heuristics


Andyyyy64:
## Context

v0.5.4 fixed the immediate #16 symptoms by treating Strix Halo / Ryzen AI MAX as AMD shared-memory APUs instead of CPU-only or 512 MB dGPU-like systems.

## Follow-up scope

- Broaden detection beyond the current name markers (`STRXLGEN`, `Radeon 8050S`, `Radeon 8060S`, `Strix Halo`, `Ryzen AI MAX`) using PCI IDs / product metadata when available.
- Replace or refine the fixed 256 GB/s bandwidth estimate when reliable system-specific data is available.
- Model Vulkan / ROCm allocation limits, UMA/BIOS configuration, and OS-reserved memory instead of relying only on `RAM * 0.80`.
- Validate whether the additive `shared_memory` JSON field needs explicit schema/docs.
- Coordinate MXFP4 / NVFP4 quant support with #27 rather than duplicating that work.
- Collect real outputs from Framework Desktop / Ryzen AI MAX users for `whichllm hardware`, `whichllm --status --top 20`, and `whichllm --json --top 5`.

## Non-goals

- Do not reopen the original #16 CPU-only / 0 tok/s regression; that was fixed and released in v0.5.4.
- Do not implement MXFP4 directly here; track it in #27.

Znuff:
```
# uvx whichllm@latest

╭───────────────────────────────────────────────── Hardware Info ──────────────────────────────────────────────────╮
│ GPU 0: AMD Radeon Graphics — 512 MB (budget 0 KB) (ROCm 6.19.14.31400100) — BW: N/A                              │
│ CPU: AMD RYZEN AI MAX+ 395 w/ Radeon 8060S — 16 cores (AVX2, AVX-512)                                            │
│ RAM: 124.9 GB                                                                                                    │
│ Disk free: 792.2 GB                                                                                              │
│ OS: linux                                                                                                        │
│ VRAM headroom: 512 MB reserved per GPU                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

`0.5.16` only lists CPU models for Strix Halo, despite the claim that it's fixed in `0.5.4`.

Andyyyy64:
Thanks for posting the output, and sorry I missed it. The 512 MB result shows that this detection path still needs work.

[...]

PR #154, opened 2026-08-09
fix: detect Strix Halo by AMD PCI id

hogeheer499-commits:
## Summary

- map AMD PCI device `1002:1586` to `Strix Halo [Radeon 8060S]`
- use the mapping for both sysfs and `lspci -mm -nn` fallback detection
- add regression coverage for both Linux probe paths

This addresses one of the PCI-ID follow-ups in #36.

## Real hardware reproduction

Tested on a Beelink GTR9 Pro with a Ryzen AI MAX+ 395 / Radeon 8060S. On current `main`, Linux exposes the GPU as generic PCI device `1586`, so `whichllm hardware` reported:

```text
GPU 0: Device 1586 — 512 MB — BW: N/A
```

With this patch, the same machine reports:

```text
GPU 0: Strix Halo [Radeon 8060S] — shared memory — BW: 256 GB/s
```

The 512 MB aperture is then normalized correctly instead of being treated as dedicated VRAM. The mapping also activates the existing Strix Halo shared-memory and bandwidth logic; it does not add a new performance heuristic.

The reproduction machine is the primary system used for the public [AMD Strix Halo Local LLM Guide](https://github.com/hogeheer499-commits/strix-halo-guide).

## Validation

- `uv run --locked ruff check .`
- `uv run --locked ruff format --check .`
- `uv run --locked pytest` (`471 passed`)
- live `whichllm hardware` check on the affected Ryzen AI MAX+ 395 system


Andyyyy64 (changes requested):
sorry this took me so long to review

thanks for including the real-hardware reproduction — the parser change and the shared-memory path look good. there's one thing blocking the merge:

`1002:1586` isn't unique to the Radeon 8060S. the current pci.ids entry is `Strix Halo [Radeon Graphics / Radeon 8050S Graphics / Radeon 8060S Graphics]`, so mapping every generic `1586` device to `Strix Halo [Radeon 8060S]` would mislabel 8050S systems.

could you use a SKU-neutral name such as `Strix Halo [Radeon Graphics]` or the canonical combined label, and update the assertions to match? it will still activate the existing Strix Halo shared-memory and 256 GB/s paths.

the Actions runs are currently `action_required` and didn't start, so we'll also need the locked lint and test checks to run after the update.

PR #159, opened 2026-08-14
Add -v as CLI synonym for --version

Sophist-UK:
## What

Add -v as CLI synonym for --version

## Why

Ease of use

## Testing

- [ ] Tests pass (`pytest`)
- [ ] New tests added (if applicable)

### DIFF
--- a/src/whichllm/cli.py
+++ b/src/whichllm/cli.py
@@ -459,6 +459,7 @@ def main(
     show_version: bool = typer.Option(
         False,
         "--version",
+        "-v",
         help="Show version and exit",
         callback=_print_version,
         is_eager=True,


Sophist-UK:
I would have add `-h` as a synonym for `--help` but I couldn't work out how.

MohammedAlkindi (commented):
Ran this on Windows 11 from the PR head (`946e6d0`):

- `whichllm --version` → prints and exits 0 (shows `unknown` from a source tree, as expected without installed metadata)
- `whichllm -v` → identical behaviour
- `whichllm -v` on current `main` → usage error, exit 2 — so the flag is genuinely new, and nothing else in `cli.py` claims `-v` (`verbose` exists only as an internal kwarg, not a CLI option), so there's no collision for typer to fight over.

On the question in your comment — `-h` as a synonym for `--help` is a one-liner in typer, on the app constructor rather than per-option:

```python
app = typer.Typer(
    context_settings={"help_option_names": ["-h", "--help"]},
    ...
)
```

I tried exactly that against this branch and `whichllm -h` prints the full usage. It would fit naturally in this PR since it's the same "short synonyms for the obvious flags" idea, but it also works as a follow-up.

Sophist-UK:
Do you want me to add the code for `-h` then?

MohammedAlkindi:
That's really Andyyyy64's call, but if it were mine I'd keep this PR as just `-v` so it stays easy to merge, and do `-h` as a separate PR. If you open one I'll run it on Windows the same way.

Andyyyy64:
Sorry I left this waiting. I checked that `-v` behaves the same as `--version`, and all 95 CLI tests pass locally.

GitHub will no longer rerun these checks because the runs are over a month old. Could you update the branch against current main so fresh CI runs can start? Please keep this PR limited to `-v`.

Andyyyy64 (approved):
Thanks for the small CLI improvement. I checked that `-v` matches `--version`, and all 95 CLI tests pass locally.

The old CI runs have expired. I'll update the branch against main and require fresh checks to pass before merging.

PR #162, opened 2026-08-15
refactor(cli): split command helpers into focused modules

RaghavRD:
Final #41 refactor slice for `src/whichllm/cli.py`: move command bodies, validation, hardware overrides, model lookup, runtime script generation, and shared CLI utilities into focused modules.

`whichllm.cli` stays the Typer entry point (`whichllm = "whichllm.cli:app"`) and keeps every command signature and its help text. It no longer re-exports internal helpers; tests import them from their new modules.

## What

- Move command execution bodies into `whichllm.cli_commands`
- Move model loading, search, GGUF selection, dependency resolution, and chat script generation into `whichllm.cli_models`
- Move shared console, async runner, version, and fetch-error helpers into `whichllm.cli_shared`
- Move validation, memory budgets, GPU overrides, ranking filters, and publish-date backfill helpers into `whichllm.cli_validation`
- Keep `whichllm.cli` to the Typer app and command signatures only (no helper re-exports, no test-only aliases)
- Update `tests/test_cli.py` to import helpers from their new modules and patch `whichllm.cli_commands` directly
- Restore the full `upgrade` docstring so `whichllm upgrade --help` is unchanged

## Why

This completes the remaining large core-module split for #41. After the previous constants, display, fetcher, benchmark, and ranker splits, `src/whichllm/cli.py` was the last large core module left.

## Testing

- [x] CLI tests pass (`95 passed`)
- [x] Full test suite passes (`484 passed`)
- [x] Compileall passes
- [x] Ruff check passes
- [x] Ruff format check passes
- [x] `whichllm upgrade --help` shows the full description again

## Notes

`uv run` is currently blocked locally because this repo now requires `uv ==0.11.33` while my local `uv` is `0.11.17`, so I ran verification through the existing `.venv`:

- `.venv/bin/python -m compileall -q src/whichllm`
- `.venv/bin/python -m pytest -q tests/test_cli.py`
- `.venv/bin/python -m pytest -q`
- `.venv/bin/python -m ruff check .`
- `.venv/bin/python -m ruff format --check .`


MohammedAlkindi:
Ran the branch rather than read it. The split is behavior-preserving in the executable sense, with one thing worth restoring before merge.

Setup: clean clone, Windows 11, Python 3.14.7, `pip install -e .`, merge-base `ea32ed2` (== current main), PR head `e0f66ed`.

- **Full suite, both sides:** `3 failed, 481 passed` at the merge-base and `3 failed, 481 passed` at `e0f66ed`, and the failing sets are identical — all three are `tests/test_asahi_detection.py` symlink-privilege failures (`WinError 1314`) that fail the same way on main, i.e. my environment, not this PR. **No new failures.**
- **Import shim holds:** everything `tests/test_cli.py` imports from `whichllm.cli` still resolves (95 passed), and all four new modules import clean. The `whichllm.cli` namespace no longer re-exports incidental imports (`sys`, `re`, `asyncio`, `Console`, `HardwareInfo`, …); nothing in the repo reaches through those, so nothing breaks.
- **Moves:** comparing all 37 functions that left `cli.py` against their merge-base source: none missing, 20 byte-identical, 6 command bodies deliberately restructured into the keyword-only `*_command` form (decorators staying in `cli.py`), and 11 helpers whose only difference is trimmed docstrings/comments.
- **CLI surface:** `--help` for the top level and for `plan`/`run`/`snippet`/`hardware` is identical on both sides. `upgrade --help` is not:

```diff
 --- merge-base
 +++ e0f66ed
-  For each GPU passed on the command line, simulate a system with the same
-  CPU/RAM but that GPU, run the ranker, and show the best-N models you'd
-  be able to run. Useful for answering "is upgrading from a 3090 to a 4090
-  worth it?" — the table shows the quality jump and the speed jump for
-  each option.
```

That last point is the one ask: the move trimmed several multi-paragraph rationale docstrings (`_merge_model_eval_benchmarks`'s deprecation history, `_auto_min_params_for_profile`, `_parse_size_tokens`, …) and the `upgrade` command's docstring — and Typer renders that one as the help text, so the deletion is user-visible. The code moves themselves look right; restoring the docstrings, or at minimum `upgrade`'s, would make this a pure move.

Andyyyy64 (changes requested):
Sorry I left the final part of this refactor waiting. I previously asked you to preserve the old imports, so I should be clear that this is a change in direction from my earlier request.

Please move the tests to import internal helpers from their new modules rather than adding re-exports in `cli.py`. Keep the CLI entry point and command behavior intact.

Mohammed also caught the missing `upgrade --help` text. Please restore that user-facing description. There is no need to restore internal comments just to make the move byte-for-byte identical.

RaghavRD:
@Andyyyy64 No problem, and thanks for being clear about the change in direction. Updated in e3121bb:

- `tests/test_cli.py` now imports helpers from `cli_models`, `cli_validation`, `cli_shared`, and `whichllm.models.artifacts`; only `app` comes from `whichllm.cli`.
- Removed the helper re-exports and the `_find_gguf_variant` / `_resolve_ranked_gguf_for_run` aliases from `cli.py`. It now holds just the Typer app and command signatures.
- `plan`/`run`/`snippet` no longer pass helpers in as arguments; `cli_commands` imports them directly, and the tests monkeypatch `whichllm.cli_commands._load_models` / `_generate_chat_script`.
- Restored the full `upgrade` docstring, so `whichllm upgrade --help` matches `main` again. I left the internal comments as they are, as you suggested.

Entry point and command behavior are unchanged. Full suite: 484 passed; ruff check and format check pass.

@MohammedAlkindi thanks a lot for running the branch and comparing all the moved functions. That's how the missing `upgrade --help` text got caught.

Andyyyy64:
Thanks for removing the re-exports and restoring the help text. The requested changes check out: all 484 tests pass on your branch, and the command help matches the pre-split CLI.

Since then, #164 has added the exact Repo ID lookup to `plan`. Could you rebase onto current main and carry that path and its tests into the extracted modules? The remaining work can stay focused on that integration; there's no need for more cleanup.

PR #164, opened 2026-08-18
fix(plan): fetch exact Hugging Face repo IDs

### LINKED ISSUE #163: Add a specific repo ID with a command
### Problem

Hi, I was trying to check with whichllm for a specific repoID in huggingface (**ilsp/Llama-Krikri-8B-Instruct** - a trained for Greek model) with '**_whichllm plan "ilsp/Llama-Krikri-8B-Instruct"_** ',  and I kept getting "_No model found matching 'ilsp/Llama-Krikri-8B-Instruct'._"

Would it be possible to add a command f.e. "--RepoID" in order to get a specific RepoID not included by default?

### Proposed Solution

The same way that whichllm downloads data for the well known models maybe it could be adopted to match the requirements of the "--RepoID" command.

### Alternatives Considered

_No response_

hannibal-lee:
## What

- Fetch an exact Hugging Face repository when `whichllm plan owner/repo` cannot find it in the cached model catalog.
- Skip the bulk model catalog fetch for exact Repo ID queries.
- Reuse the existing model metadata parser and respect `HF_ENDPOINT`.
- Add clear errors for missing, private or gated repositories, network failures, and insufficient model metadata.
- Call the new fetch directly from `models.hf` instead of adding another compatibility wrapper to `models.fetcher`.
- Keep directly fetched repositories out of the shared ranking cache.

## Why

`whichllm plan` documents support for Hugging Face Repo IDs, but it previously searched only the limited cached catalog. Valid repositories outside that catalog returned `No model found`.

This particularly affected `ilsp/Llama-Krikri-8B-Instruct`, which exists on Hugging Face but does not appear in the default model listing.

Fixes #163

## Testing

- [x] Tests pass (`pytest`) — 509 passed
- [x] New tests added
- [ ] Tested on real hardware (not hardware-related)

## Notes

Verified against the real Hugging Face API with:

```bash
whichllm plan ilsp/Llama-Krikri-8B-Instruct --json
```

Hugging Face answers `401` (`{"error":"Invalid username or password."}`) for repositories that do not exist, not `404`, so a mistyped Repo ID now reports "Repository not found" instead of an access error.


Andyyyy64 (changes requested):
Sorry this has been waiting. The exact Repo ID lookup addresses #163, and keeping it out of the shared catalog cache makes sense.

Could you call the new function directly from `models.hf` and update the tests to patch that module? The new wrapper in `models.fetcher` temporarily changes a module-level function just to preserve the old patching path. I want to avoid extending that compatibility layer.

hannibal-lee:
Thanks for the review — both points are addressed:

- `plan` now calls `fetch_model_by_id` directly from `whichllm.models.hf`, so `models.fetcher` is untouched: no new wrapper and no new `__all__` entry. The PR no longer modifies that file at all.
- The CLI tests patch `whichllm.models.hf` instead. The function's own unit test moved to a new `tests/test_hf.py`, where it patches `hf.get_with_retries`; `tests/test_fetcher.py` is unchanged.

While verifying the error paths against the real API I found that Hugging Face returns `401` with `{"error":"Invalid username or password."}` for repositories that do not exist — not `404` — so the old 404 branch never fired and a typo was reported as "private or gated". `_raise_repo_fetch_error` now treats that response as "Repository not found on Hugging Face" and keeps the access message for restricted responses. Covered by three new tests (JSON body, plain-text body, restricted body).

I also rebased onto current `main` (no conflict with #167's frontier list) so fresh CI has run, and tidied the commit message, which had a literal `\n\n`.

Andyyyy64 (approved):
Thanks for updating this. The new lookup now calls `models.hf` directly, and the error message allows for repositories that are private or unavailable rather than claiming we can distinguish every case.

All 509 tests pass locally. I also checked `plan ilsp/Llama-Krikri-8B-Instruct --json` against the public API with an empty cache, and it returns the expected model. Looks good to merge.

PR #166, opened 2026-08-19
docs: fix broken star history chart

FaintFlower:
The star history chart in the README currently renders blank because the previous provider is affected by GitHub stargazer API restrictions. This switches the chart to a working alternative so the repository's star history is visible again.


Andyyyy64 (approved):
Sorry for the wait, and thanks for fixing this. The old image URL returns 403 in my check, while the replacement renders the repository's star history correctly in the browser. All CI checks pass. Looks good to merge.

PR #167, opened 2026-08-21
feat(models): cover newer Qwen releases and widen the math profile

### LINKED ISSUE #24: Improve coverage for newer Qwen models and task profiles
### Problem

Some newer model families are present in the cache but do not rank well because metadata or benchmark evidence is missing. Examples seen locally: `Qwen/Qwen3.6-35B-A3B` is not detected as MoE, and `Qwen/Qwen3.5-9B` / newer coder models have weak or missing benchmark evidence.

This also affects task profiles: `--profile math` is currently very sparse, and `--profile coding` still falls back to older Qwen2.5-Coder models quickly.

### Done when

- `Qwen/Qwen3.6-35B-A3B` is parsed with correct MoE/active-parameter metadata or an equivalent curated override.
- Newer Qwen3.5/Qwen3.6 coder/math-relevant models have reasonable direct, inherited, or curated benchmark evidence.
- `--profile coding` and `--profile math` return current, defensible recommendations on common simulated GPUs.
- Tests cover the new metadata and ranking behavior.

devangpratap:
## What

- Add the Qwen3.5 / 3.6 / 3.8 releases and the current coder line to the frontier fetch: `Qwen3.8-27B`, `Qwen3.6-35B-A3B`, `Qwen3.5-27B`, `Qwen3.5-35B-A3B`, `Qwen3.5-122B-A10B`, `Qwen3.5-9B`, `Qwen3.5-4B`, `Qwen3-Coder-Next`, `Qwen3-Coder-480B-A35B-Instruct`.
- Map them to their AA display names and add curated snapshot entries for the offline path.
- Pin `Qwen3-Coder-Next` at 3B active params.
- Add a `qwen3.8` lineage row.
- `--profile math` keeps general reasoning models, not just repos with "math" in the name.

## Why

`Qwen3.6-35B-A3B` MoE metadata was already correct via the `A*B` naming hint (35B total, 3B active, checked against the live HF payload). The bug was that none of the 3.5/3.6 line is in HF's top-300 by downloads, so it was never fetched and never parsed. Test added to keep that path covered.

`Qwen3-Coder-Next` needed a real override. Its config is 512 experts, top-10, so the expert-fraction estimate gives ~33B active against a model card of 80B total / 3B active. VRAM and tok/s were both wrong.

Without AA mappings the new repos could only inherit a line-interpolated score. Raw values are from the live index (Intelligence Index v4.3, read 2026-09-14), highest variant per model, same as `fetch_aa_index_scores`. Two existing hand estimates were off enough to reorder neighbours offline, so I replaced them with measured values: `Qwen3.6-27B` 32.0 to 21.9, `Qwen3-Coder-30B-A3B-Instruct` 19.7 to 9.6 (the max-overlay meant live could never correct it). Nothing else in the snapshot changed and I did not touch `fetch_aa_index_scores`, that is #101 / #173.

AA moved to v4.3 in September, which lowered scores across the board. The entries this PR does not touch keep their 2026-06-29 values, which are on the older, higher scale, so offline the refreshed Qwen entries sit lower against them than they do live. Kept out of scope here, the snapshot comment and docstring note the mixed dates.

No `qwen3.8` lineage pattern meant those repos fell to the generic `qwen3` row and scored a smaller generation bonus than 3.5 and 3.6. The fine-tunes are already in the cache, so they were being demoted against older models.

`--profile math` returned two candidates on a 24GB card, the better being a 1.5B model from 2024. Math ability comes from general reasoning models now, so the filter keeps specialists eligible and only drops coding and vision. `general` is unchanged, `_matches_profile` has one caller.

## Testing

- `pytest` 495 passing, 11 new in `tests/test_newer_qwen_coverage.py`. No network, no fixtures.
- `ruff check` / `ruff format --check` / pre-commit clean.
- Ranked the real cache on a simulated Linux + RTX 4090 24GB, live AA reachable (run on 2026-08-21, before the snapshot refresh):
  - `math` before: `Qwen2.5-Math-1.5B` and a 707M community model, whole list. After: `Qwen3.8-27B`, `Qwen3.6-27B`, `gemma-4-31B-it`, `Qwen3.6-35B-A3B`, `Qwen3.5-27B`.
  - `coding` before: `Qwen3-Coder-30B-A3B-Instruct` plus four mirrors of it. After: `Qwen3-Coder-30B-A3B-Instruct`, `Qwen3-Coder-Next`, `ManniX-ITA/Qwen3.6-27B-A3B-Coder-MTP-GGUF` (third-party conversion), `Qwen2.5-Coder-32B-Instruct` on direct evidence.

## Notes

No math-specific score axis yet, so `math` is the general ranking minus coding and vision specialists. LiveBench publishes math subscores if you want that as a follow-up.

Closes #24


Andyyyy64 (changes requested):
Sorry I left this waiting, and thanks for working through the newer Qwen coverage. All 495 tests pass locally, and the current Hugging Face metadata supports the model IDs and the Coder Next active-parameter correction.

It looks like the AA data has moved since you collected the snapshot. I now get about 33.9 as the highest Qwen3.8 27B entry, compared with the 52.0 stored in this patch. Could you refresh the added and changed entries against the current page and update the snapshot dates to match? The table and docstring still say June, while the additions were collected in August.

Please keep this focused on the model coverage and snapshot update. The live-versus-stored merge change is in #173, and the family-grouping question is tracked in #171.

devangpratap:
Great to see you back in action. I've refreshed the added and changed entries against the current page (Qwen3.8 27B is now 33.9, the highest variant) and updated the snapshot name, docstring, and dates to 2026-09-14.

One heads up: AA moved to Intelligence Index v4.3 in September, so the entries this PR doesn't touch are still on the older, higher scale. I kept this PR scoped, but figured it's useful context for #173.

Let me know if it needs any further revisions.

Andyyyy64 (approved):
Thanks for refreshing these and for pointing out the scale change. I checked the 13 updated entries against the current AA page, and they match to one decimal place. All 495 tests pass locally too.

The mixed snapshot dates are now documented. I'll keep the remaining older-scale entries separate from this PR so they don't hold up your model-coverage work. Looks good to merge.

devangpratap:
Thanks for the merge. Is there anything else you'd like me to pick up?

PR #168, opened 2026-08-24
feat: display model disk size column in recommended models table

RBCs-lang:
## Why

Currently, the default ranking table shows hardware configuration and runtime memory requirements (`Fit / VRAM` or
`Fit / RAM`), but does not display the model weight footprint. Users need visibility into
how much storage/download size each recommended model requires to plan their disk space and downloads.

## Outcome

- Adds a **Weights** column immediately before `Fit / VRAM` (or `Fit / RAM`) in the Recommended Models table when status is displayed.
- Displays the exact file size from the resolved GGUF variant when available (e.g. `15.0 GB`).
- For non-GGUF or unquantized candidates where exact file size is not directly stored, falls back to `estimate_weight_bytes` and marks the value with a tilde prefix (e.g. `~3.3 GB`).
- Adds a footnote (`Weights: ~ = estimated size`) whenever estimated values appear in the ranking.
- Keeps `pyproject.toml` and package dependencies unchanged.

## Changes

- **`src/whichllm/output/ranking.py`**:
- Added `Weights` column (`width=8`, `justify="right"`) in `display_ranking` when `show_status` is true.
- Formatted cell values with exact GGUF file size when present, or estimated weight with a `~` prefix.
- Added a footnote line clarifying the `~` marker for estimated weights.
- **`tests/test_output_formatting.py`**:
- Added test coverage verifying the `Weights` column header, actual size formatting, estimated size marker (`~`), and footnote display.

    ## Validation

    - [x] Full pytest suite passes (`pytest` — 485 passed).
    - [x] Linter passes cleanly (`ruff check .`).
    - [x] Verified table rendering across standard and narrow terminal widths.


Andyyyy64 (changes requested):
Sorry for the late review. Showing model size is useful, but please keep the uv version pin unchanged in this PR.

The new value comes from `estimate_weight_bytes`, so it is not always the actual download size. Could you label it as weights and mark estimated values, while using the file size when available? Please also check the table at a narrow terminal width.

RBCs-lang:
Thanks for the review @Andyyyy64! I've updated the PR accordingly:- 

1. `uv` version: Reverted `pyproject.toml` back to `==0.11.33`.
2. `Weights` column & estimations:
- Relabeled the column to `Weights`.
- Uses the actual GGUF file size when available.
- Falls back to `estimate_weight_bytes` with a `~` prefix (e.g. `~3.3 GB`) and adds a `Weights: ~ = estimated size` footnote when estimated values are present.
3. Terminal width: Tested table behavior across standard 80-column and resized narrow widths.
4. Rebase: Cleanly rebased onto the latest `main`.

Marking ready for review!

Andyyyy64 (changes requested):
Thanks for updating the column and restoring the uv pin. There's one remaining case: synthetic GGUF variants already have an estimated number in `file_size_bytes`, so the current check displays it without `~`. I reproduced this with a model that has no GGUF files.

Could you preserve whether the size came from file metadata or an estimate, and use that when rendering the column? Please cover synthetic variants and missing file-size metadata in the regression test too.

RBCs-lang:
Thanks for the review @Andyyyy64! I've updated the PR accordingly:-

1. `GGUFVariant` size provenance:
   - Added `is_estimated: bool = False` to `GGUFVariant` to preserve whether the size came from file metadata or an estimate.
   - Set `is_estimated=True` for synthetic variants generated for official repos (`_synthesize_variants_for_official_repo`).
   - Set `is_estimated=True` when sibling file size metadata is missing or non-positive in Hugging Face Hub metadata (`_extract_gguf_variants`), while keeping `is_estimated=False` for actual metadata sizes.
   - Preserved `is_estimated` across cache serialization and deserialization in `models_to_dicts` and `dicts_to_models`.

2. `Weights` column rendering & footnote:
   - Updated `_format_weights` to check `not variant.is_estimated`: displays exact metadata sizes cleanly without `~` and marks all estimated values (synthetic variants, missing file-size metadata, non-GGUF models) with `~`.
   - Linked the `Weights: ~ = estimated size` footnote directly to `_format_weights`, ensuring it is only displayed when estimated values are present in the table.

3. Regression & unit test coverage:
   - Added tests in `tests/test_output_formatting.py` covering:
     - Synthetic GGUF variants (verifying `~` prefix and footnote display).
     - Missing file-size metadata variants (verifying `~` prefix and footnote display).
     - Exact file metadata sizes (verifying clean formatting without `~` and absence of footnote).
     - Direct unit tests for `_format_weights`.
     - Flag assignment in `_synthesize_variants_for_official_repo` and `_extract_gguf_variants`.
   - Added cache roundtrip test in `tests/test_fetcher.py` verifying `is_estimated` persistence.

4. Validation:
   - Full test suite passes (`pytest` — 503 passed).
   - Linters and code formatter pass cleanly (`ruff check .`, `ruff format --check .`).

Andyyyy64:
Thanks for following up on this. The cache roundtrip test is helpful here, since the estimated-size flag needs to survive a reload too. I'll recheck the updated branch against the synthetic GGUF case that failed before.

Issue #20, opened 2026-05-16
LM Studio


ruze00:
### Problem

Duplication of models already on the system. 

### Proposed Solution

It would be awesome if it could integrate with models already on the system from LM Studio instead of pulling independently.

Andyyyy64:
Thanks for the suggestion, and sorry this went unanswered.

This is still not implemented. whichllm can recommend and run GGUF models through Hugging Face, but it does not currently scan LM Studio’s local model library or mark recommendations as already installed.

I agree with the problem, though. If someone already has GGUF models installed through LM Studio, whichllm should not make them download the same model again just to act on a recommendation.

I think the first version should be read-only:

- detect common LM Studio model directories on macOS, Windows, and Linux
- allow a custom path for non-default libraries
- match local GGUF files to model IDs / quant types where possible
- show an "already installed" or "local match" hint in the recommendation output

I would avoid changing or managing LM Studio’s library from whichllm. Detecting what is already there is the safer first step.

I will keep this open as the tracking issue for LM Studio local-library integration.

PR #169, opened 2026-08-25
feat: detect local LM Studio GGUF models

hannibal-lee:
## Summary

- scan only the current `~/.lmstudio/models` library automatically, with repeatable `--lm-studio-path` options for explicit legacy or custom library roots
- match installed models only when the resolved Hugging Face repository and complete artifact path both match
- require every part of split GGUF artifacts and reject missing, broken, or unreadable files
- report missing, non-directory, and unreadable explicit paths before model or benchmark network work begins
- mark verified local matches in Rich and Markdown output and expose `local_match` / `local_path` in JSON

## Testing

- `.venv/bin/pytest -q` (`535 passed`)
- `.venv/bin/ruff check .`
- `.venv/bin/ruff format --check .`
- `git diff --check`
- `uv build --no-config --wheel`

## Notes

This remains the read-only first version discussed in #20. It does not modify or manage LM Studio files, and it does not yet make `whichllm run` launch the local artifact directly.

Addresses #20


hannibal-lee:
Closing because this PR was opened from the wrong change set.

Andyyyy64:
Sorry I left this waiting. I saw your comment about closing the PR because it came from the wrong change set, but it is still open. Would you like to continue with this version or withdraw it? I'll hold off on the review until that is clear.

hannibal-lee:
Thanks for checking in, and sorry for leaving this ambiguous. The "wrong change set" comment was my own false alarm — this branch is a single commit on top of `ea32ed2` and the diff is exactly the LM Studio change set described above; nothing was ever rewritten. I'd like to continue with this version.

I've rebased onto current `main` so fresh CI has started. The only conflict was `CHANGELOG.md` (merged with #167's entry); the LM Studio change itself is unchanged, and this stays the read-only first step from #20 — detection only, no library management.

Andyyyy64 (changes requested):
Thanks for clearing that up. I'm happy to review this as the read-only first step.

I reproduced a false match: a different filename in the same repo is marked installed when its quantization matches. Please match the resolved repo and artifact rather than falling back to quantization alone or a filename without a repo. Missing split-file parts and broken symlinks should not count as an installed model either.

Please keep automatic discovery to the current library location and use `--lm-studio-path` for other layouts, including older installations. If an explicitly supplied path cannot be read, please report that rather than silently treating it as an empty library.

hannibal-lee:
Thanks for the review. I’ll tighten matching to require the resolved repo and exact artifact, verify all split parts and symlinks, limit automatic discovery to the current LM Studio library, and report unreadable explicit paths. I’ll add regression tests and push an update.

hannibal-lee:
Implemented the requested changes and rebased onto current `main`:

- matching now requires the resolved repository and exact repository-relative artifact path; the quantization-only and filename-without-repo fallbacks are removed
- split GGUFs are marked installed only when every numbered part is present and readable
- broken file symlinks are ignored, while symlinks that resolve to readable files remain valid
- automatic discovery is limited to `~/.lmstudio/models`; legacy and custom roots require `--lm-studio-path`
- missing, non-directory, or unreadable explicit paths now produce a CLI error before network work

I added regression coverage for the reproduced same-repo/same-quant/different-filename false match, complete and incomplete splits, valid and broken symlinks, explicit path failures, and current-vs-legacy discovery. Local verification: 535 tests passed, Ruff check/format passed, and the wheel build succeeded.

Andyyyy64:
Thanks for working through these cases and rebasing. Sorry I've kept you waiting on the review. I'll check the updated matching against the original false match, then go through the split-file and path handling.

PR #173, opened 2026-09-14
Use live AA scores even when they decrease

Andyyyy64:
Live AA scores currently replace stored scores only when higher. The parsers also discard zero and negative raw scores, so a successful fetch can leave an outdated score in place.

Accept finite raw scores and let mapped live values replace stored AA values. Models absent from the response retain their stored scores. This keeps the normalization bounds and cross-source merge policy unchanged.

495 tests pass locally on Python 3.13, along with Ruff and format checks. The added cases cover decreases, zero, negative raw scores, and invalid payload values. Existing benchmark caches expire normally; `--refresh` applies the new results immediately.

Refs #101


devangpratap (commented):
Ran this branch against the current AA page and noticed something that might not be intended.

`AA_NAME_TO_HF_IDS` maps `"DeepSeek V3"` to both `DeepSeek-V3` and `DeepSeek-V3-0324`, and `"GLM-4.5"` to both `GLM-4.5` and `GLM-4.5-Air`. AA lists `DeepSeek V3 0324` (slug `deepseek-v3-0324`) and `GLM-4.5-Air` (slug `glm-4-5-air`) as their own rows, but those names don't map, so each repo gets its sibling's score instead of its own:

| Repo | Stored | This branch | Its own AA row |
| --- | --- | --- | --- |
| `deepseek-ai/DeepSeek-V3-0324` | 44.5 | 41.6 (from `DeepSeek V3 (Dec '24)`) | 43.5 |
| `zai-org/GLM-4.5-Air` | 58.1 | 48.0 (from `GLM-4.5 (Reasoning)`) | 45.5 |

(normalized, current bounds; AA marks both GLM rows as estimated)

On main the lower live value never replaced the stored one, so this didn't show. With live replacing stored it now reaches the rankings. Is that intended? If not, giving each checkpoint its own entry fixes it:

```python
"DeepSeek V3": ["deepseek-ai/DeepSeek-V3"],
"DeepSeek V3 0324": ["deepseek-ai/DeepSeek-V3-0324"],
"GLM-4.5": ["zai-org/GLM-4.5"],
"GLM-4.5-Air": ["zai-org/GLM-4.5-Air"],
```

Happy to send that as a follow-up if that's easier.

Andyyyy64:
Thanks for catching the checkpoint mix-up, and for offering to help. Could you send a small PR that gives DeepSeek V3 and V3-0324 their own AA entries, and does the same for GLM-4.5 and GLM-4.5-Air?

A regression test showing that each checkpoint receives its own row would be helpful. I'll handle the zero-score aggregation fix separately, so there's no need to take on the merge-policy change.

devangpratap:
Opened #181 for this. Each of the four checkpoints now maps to its own AA display name, with a regression test pinning each one to a single id.

I checked the new keys against AA's model pages rather than guessing: "DeepSeek V3 0324" and "GLM-4.5-Air" both match verbatim. No snapshot values changed, and I left the zero-score aggregation and the merge policy alone as you said.

PR #181, opened 2026-09-19
fix(aa-index): give each checkpoint its own AA entry

devangpratap:
Follow-up to the checkpoint mix-up in #173.

`AA_NAME_TO_HF_IDS` aliased two checkpoints onto one display name each:

    "DeepSeek V3": ["deepseek-ai/DeepSeek-V3", "deepseek-ai/DeepSeek-V3-0324"]
    "GLM-4.5":     ["zai-org/GLM-4.5", "zai-org/GLM-4.5-Air"]

Both ids in each pair took the base model's index, and a live row for the
sibling was dropped because no display name matched it. These are separate
models, not quantizations: `parameters.py` has GLM-4.5 at 32B active against 12B
for GLM-4.5-Air, and livebench scores them separately (58.0 vs 52.0). aider
lists `deepseek-v3` and `deepseek-v3-0324` as separate rows.

I checked the new keys against AA's own model pages: "DeepSeek V3 0324" (index
10, against 8 for "DeepSeek V3 (Dec '24)") and "GLM-4.5-Air" (index 11). Both
match verbatim, and the parenthetical on the plain V3 label is stripped by
`_canonical_name`, so it still resolves to the base id.

Left alone: the quantization pairs that legitimately share a row, GLM-5/GLM-5-FP8,
GLM-5.1/GLM-5.1-FP8 and Nemotron 3 Nano BF16/FP8. Two other entries are not
quantization pairs, "Kimi K2" (Instruct and Base) and "Command R+" (two dated
repos). The same reasoning applies, but I kept this PR to what you asked for.
Happy to send a second.

No snapshot values changed. DeepSeek-V3 and V3-0324 stay at 10.4 and the GLM-4.5
pair at 19.5, since those came from the shared row and I have no separate
measurement on that scale. GLM-4.5-Air cannot lose coverage either way: the live
fetch overlays onto the curated fallback, so with no Air row it keeps its
snapshot value rather than the base model's.

### Testing

Two tests in `tests/test_aa_index.py`: one pinning each display name and the
canonical map to a single id, one driving the live fetch with both GLM-4.5 rows
present and asserting the ids get different scores. Against main the second fails
with `assert 58.1 != 58.1`. Full suite 518 passing, ruff check and format clean.


Andyyyy64:
Thanks for taking this on and checking the names against AA's pages. I'll review this first, then check it together with the zero-score fix in #173.

Thanks for flagging Kimi K2 and Command R+ too. Let's keep those separate so they don't hold up these four checkpoints.

